Issue #16, opened 2017-10-01
Improve installation file


vrom911:
If `~/.local/bin` doesn't exist then add it.

PR #20, opened 2017-10-06
[#16] Create ~/.local/bin during install if it does not exist

axocomm:
This just adds a quick check for the existence of `~/.local/bin` during installation and creates it if necessary.


vrom911:
@axocomm  Thank you very much! Great work :+1:

Issue #17, opened 2017-10-01
Ability to add own default values during installation


vrom911:
Now the script uses my data in default values, that  is written in code.
Rather to dive into the script and change values there by hands I'd like to add the opportunity to change them automatically during the installation.

axocomm:
I'm looking to get more familiar with Haskell and I'd love to give this one a try if you don't mind!

If so, may I ask if you're referring to the installation shell script (i.e. populating these values in `hs-init.hs` before copying to `~/.local/bin`) or having each run of `hs-init` allow specification, either by command line options or perhaps a configuration file, of these details?

vrom911:
@axocomm  First of all thank you for your interest in this tool!

To clarify the issue, my plan is to add this actions to the installation script (`install`), so during the installation process you can somehow input `defaultOwner`, `defaultName`, `defaultEmail`, `defaultLicense` and `defaultGHC` and the script should replace this data directly in `hs-init` file. After installation `hs-init` should have data received on previous step  in corresponding fields. 

Adding command line option to correct this data is great idea! It will be really useful, but it's different issue, I guess. We can start with this feature now.

axocomm:
Happy to help!

So I've got a quick solution that simply does a few `sed`s to `hs-init.hs` before moving it to `~/.local/bin`. I just have a few clarifying questions:

* For actually specifying these values, were you thinking more along the lines of a configuration file:
    ``` shell
    DEFAULT_NAME=foobar
    DEFAULT_EMAIL=foo@bar.com
    ...
    ```
    or going with command line options/environment variables?

* Should these all be required? For year and license, those could conceivably default to `date +'%Y'` and MIT respectively, but what about the owner; email; license; etc.? Perhaps just let a neglectful user give you the credit 😛 ?

vrom911:
I think that all data should be asked interactively during work of installation script. No need for command line arguments for that. I would rather not to have any configurations files at users machine, cause I want to keep it very simple.

Also I guess none of the fields are required, and if they are not  entered during the installation we just put some template data in there. The idea of this project is that this data is asked every time during the work of the tool anyway, cause it's not permanent data, I want more flexibility and give users the opportunity to create projects with different data each time. Default values are just for making the process quick and smooth if you're using this data every time.

And thank for your comment about time. I will make a function for that directly in `hs-init.hs` which will take the current year, no need to input it during the installation :slightly_smiling_face:

PR #21, opened 2017-10-07
[#17] Prompt for default settings

axocomm:
These changes update the install script to prompt for defaults for owner, name, and email address.

### DIFF
--- a/hs-init.hs
+++ b/hs-init.hs
@@ -49,22 +49,22 @@ import           System.Process        (callCommand, readProcess, showCommandFor
 -----------------------
 
 defaultOwner :: Text
-defaultOwner = "vrom911"
+defaultOwner = "{{DEFAULT_OWNER}}"
 
 defaultName :: Text
-defaultName = "Veronika Romashkina"
+defaultName = "{{DEFAULT_NAME}}"
 
 defaultEmail :: Text
-defaultEmail = "vrom911@gmail.com"
+defaultEmail = "{{DEFAULT_EMAIL}}"
 
 defaultLicense :: Text
-defaultLicense = "MIT"
+defaultLicense = "{{DEFAULT_LICENSE}}"
 
 defaultGHC :: Text
-defaultGHC = "8.0.1"
+defaultGHC = "{{DEFAULT_GHC}}"
 
 defaultYear :: Text
-defaultYear = "2017"
+defaultYear = "{{DEFAULT_YEAR}}"
 
 --------------------------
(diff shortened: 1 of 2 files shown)


vrom911:
Amazing! Thank you very much for your work. 
I will review and test it now and will write some comments :slightly_smiling_face:

hs-init.hs
```diff
@@ -49,19 +49,19 @@ import           System.Process        (callCommand, readProcess, showCommandFor
...
-defaultEmail = "vrom911@gmail.com"
+defaultEmail = "{{DEFAULT_EMAIL}}"
 
 defaultLicense :: Text
-defaultLicense = "MIT"
+defaultLicense = "{{DEFAULT_LICENSE}}"
 
 defaultGHC :: Text
-defaultGHC = "8.0.1"
+defaultGHC = "{{DEFAULT_GHC}}"
```
vrom911: Oops! Now I see that it's not such a good idea to ask user write `DEFAULT_LICENSE` and `DEFAULT_GHC`. It can only be chosen from the list of options, but it's hard to maintain in the installation script. My bad!
So no need to change `defaultLicense` and  `defaultGHC`:
 ```haskell
defaultLicense :: Text		
defaultLicense = "MIT"		
  		  
 defaultGHC :: Text		 
 defaultGHC = "8.0.1"
```

File: hs-init.hs [hunk 2]
```diff
@@ -49,19 +49,19 @@ import           System.Process        (callCommand, readProcess, showCommandFor
 -----------------------
 
 defaultOwner :: Text
-defaultOwner = "vrom911"
```
vrom911: Can we keep my data in the file and change it only if the users input theirs own default values?
I'm using `stack template` to create projects and the `{{smth}}` structure creates conflicts and the tool fails.. So I want some default valid data in this variables.

axocomm: Sure!

vrom911:
Thank you very much! This works perfect for me :+1:

vrom911:
@axocomm , I just checked the script with [`shellcheck`](https://github.com/koalaman/shellcheck) and that's what I got:
```bash
In install line 37:
  IFS= read -p "$prompt" $dest <&2
  ^-- SC2162: read without -r will mangle backslashes.
                         ^-- SC2086: Double quote to prevent globbing and word splitting.
```
Could you please explain me what are these warnings about? I'm not an expert in this, but I think for you it will be easy to understand :slightly_smiling_face:

PR #24, opened 2017-10-08
[#17] Address shellcheck warnings

axocomm:
This updates the `read_default` function in the install script to address shellcheck warnings. Namely, passing the `-r` option to `read` and double-quoting the `$dest` variable.


vrom911:
Thank you very much! Amazing! :1st_place_medal:

Issue #22, opened 2017-10-07
Get current year automaticly


vrom911:
For the moment I have 
```
defaultYear :: Text
defaultYear = "2017"
```
I guess it's not a good idea to change `defaultYear` manually on Jan, 1 each year instead of having the party :champagne: 
So the plan is to use `time` package to get current year on fly.

vrom911:
Implemented in PR #25

PR #25, opened 2017-10-10
[#22] Get year automatically

puhlenbruck:
I see there was already a pull request for this issue, but it appears to have some problems so I thought I would submit my own solution.

The year string was added to the ProjectData data type and as a parameter to the customizeLicense  function as it can no longer be a globally available constant.

I am fairly new to haskell, let me know if there is anything that should be improved before merge.

### DIFF
--- a/hs-init.hs
+++ b/hs-init.hs
@@ -10,6 +10,7 @@
   --package optparse-applicative
   --package text
   --package neat-interpolation
+  --package time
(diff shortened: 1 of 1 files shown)


vrom911:
@puhlenbruck , thank you very much! This works as it should. Great work :1st_place_medal:

PR #33, opened 2017-10-22
Create CODEOWNERS

luvhalvorson:


ChShersh:
@luvhalvorson Thanks for paying your attention to this project! But the issue wasn't about adding CODEOWNERS file to `hs-init` repository. It's about adding `CODEOWNERS` to project which is generated by `hs-init` tool.

@vrom911 Do you want to have CODEOWNERS file in your `hs-init` repository? If so, I think, it's better to set @vrom911 as code owner, not @ChShersh.

vrom911:
@luvhalvorson , thank you very much for your job!
But as @ChShersh  already mentioned, it's not quite what the issue is about. `CODEOWNERS` is really cool one, but I guess for this particular not big project with very simple architecture and not big team of reviewers it does not seem reasonable to have it.
If you're still interested in implementing this feature to the project, it would be more than welcomed. Keep in mind that I'm happy to discuss and answer any questions you have to help in understanding the issue and the project itself. I will close this PR so it won't mess up with the actual solution of the problem. So feel free to open new ones :slightly_smiling_face: ! Thanks again :+1: !

PR #42, opened 2018-01-02
[#36] Option -p/--private

### LINKED ISSUE #36: Support creation of private repositories at GitHub
Add cli option to support this.

vrom911:
@ChShersh could you please review when you have time just to be sure I'm not doing strange things in the project?

Resolves #36

### DIFF
--- a/hs-init.hs
+++ b/hs-init.hs
@@ -26,36 +26,29 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies        #-}
 
-import           Control.Monad                  (when)
-import           Data.Aeson                     (FromJSON (..), decodeStrict, withObject,
-                                                 (.:))
-import           Data.ByteString.Char8          (pack)
-import           Data.Foldable                  (fold)
-import           Data.List                      (nub)
-import           Data.Semigroup                 ((<>))
-import           Data.String                    (IsString (..))
-import           Data.Text                      (Text)
-import qualified Data.Text                      as T
-import qualified Data.Text.IO                   as T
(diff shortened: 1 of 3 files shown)


hs-init.hs
```diff
@@ -188,11 +185,13 @@ generateProject repo owner description Targets{..} = do
...
-  doGithubCommands :: IO ()
-  doGithubCommands = do
+  doGithubCommands :: Bool -> IO ()
+  doGithubCommands private = do
     -- Create the repository on Github.
     "git" ["init"]
-    "hub" ["create", "-d", description, owner <> "/" <> repo]
+    -- creates private repository if asked so.
+    let priv :: Text = if private then "-p" else ""
+    "hub" ["create", priv, "-d", description, owner <> "/" <> repo]
```
ChShersh: If you don't mind  writing `private` option at then end, you can write it like this:

```haskell
"hub" $ ["create", priv, "-d", description, owner <> "/" <> repo]
     ++ ["-p" | private]
```

hs-init.hs
```diff
@@ -154,6 +147,10 @@ generateProject repo owner description Targets{..} = do
             if github
             then decisionToBool ciFlag target
             else falseMessage target
+  privat <- let privTarget = "Private repository" in
+            if github
+            then decisionToBool privateFlag privTarget
+            else falseMessage privTarget
```
ChShersh: Looks like there's some code duplication between `ci`, `privat` and `isExe` flag. Maybe this can be refactored?

vrom911:
@ChShersh yes, true, it becomes only bigger and I would like to make great refactoring to change it, but I don't have strong opinion on how it could be improved a lot without making it very heavy and not converting it into the project.

PR #43, opened 2018-01-08
Use metavar for on-off commands

ChShersh:
Nice hack I discovered from this SO answer:

* https://stackoverflow.com/questions/48144402/how-do-i-specify-names-for-missing-commands-in-the-help-message-generated-by-opt

Help message now prints:

```
Usage: hs-init.hs PROJECT_NAME [on] [off]
```

Before it was

```
Usage: hs-init.hs PROJECT_NAME [COMMAND] [COMMAND]
```


hs-init.hs
```diff
@@ -301,12 +301,18 @@ benchmarkP d =  flag Idk d
              <> help "Benchmarks"
 
 onP :: Parser Targets
-onP  = subparser $ command "on" $
-          info (helper <*> targetsP Yes) (progDesc "Specify options to enable")
+onP  = subparser $ mconcat
+       [ metavar onC
+       , command onC $ info (helper <*> targetsP Yes) (progDesc "Specify options to enable")
+       ]
+  where onC = "on [OPTIONS]"
```
Lysxia: `command "on [OPTIONS]"` seems odd.  Shouldn't it remain `command "on"`?

ChShersh: @Lysxia Sorry, my bad :( I was in hurry =\

ChShersh: Fixed in this commit: https://github.com/vrom911/hs-init/commit/7b822eae83a3d42f9f1d2fd0467177c18bbafe9f

PR #46, opened 2018-01-18
[#45] Add AppVeyor to hs-init

### LINKED ISSUE #45: Add AppVeyor CI
Currently only Travis CI is supported. I think we can also add AppVeyor CI.

ChShersh:
Resolves #45 

I've added AppVeyor as a separate CLI flag. Maybe if `--ci` specifed we can add both Travis and AppVeyor. I'm not sure what is more convenient for users. Maybe somebody don't want to deal with Windows...

I've also performed some cosmetic changes. And added `appveyor.yml` to `hs-init`.
@vrom911 you probably need to add `hs-init` project to AppVeyor to launch builds. I don't know how `chmod +x` works on Windows and how can I launch `hs-init --help` on Windows so I just build it with `ghc` on Windows.

Here is the example of project I've created to test AppVeyor:

* https://github.com/ChShersh/appveyor-test2

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,9 @@
 # hs-init
 
 [![Build status](https://secure.travis-ci.org/vrom911/hs-init.svg)](http://travis-ci.org/vrom911/hs-init)
+[![Windows build status](https://ci.appveyor.com/api/projects/status/github/vrom911/hs-init?branch=master&svg=true)](https://ci.appveyor.com/project/vrom911/hs-init)
 [![MIT license](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/vrom911/hs-init/blob/master/LICENSE)
-[![Version 1.0.2](https://img.shields.io/badge/version-v1.0.3-fabfff.svg)](https://github.com/vrom911/hs-init/blob/master/CHANGELOG.md)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,8 +1,9 @@
 # hs-init
 
 [![Build status](https://secure.travis-ci.org/vrom911/hs-init.svg)](http://travis-ci.org/vrom911/hs-init)
+[![Windows build status](https://ci.appveyor.com/api/projects/status/github/vrom911/hs-init?branch=master&svg=true)](https://ci.appveyor.com/project/vrom911/hs-init)
 [![MIT license](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/vrom911/hs-init/blob/master/LICENSE)
-[![Version 1.0.2](https://img.shields.io/badge/version-v1.0.2-fabfff.svg)](https://github.com/vrom911/hs-init/blob/master/CHANGELOG.md)
+[![Version 1.0.2](https://img.shields.io/badge/version-v1.0.3-fabfff.svg)](https://github.com/vrom911/hs-init/blob/master/CHANGELOG.md)
```
vrom911: Here it names the version twice: it the label and in the link, so the first one should be 1.0.3 also.
Yes, I know that it's not the best way to keep version updated, but I don't know to deal with it..

ChShersh: No problems :ok_hand: I also don't know better way...

Issue #27, opened 2017-10-11
Remove `b` script


vrom911:
As it seen at the screenshot coloring is not working as it should at the `ghc 8.2.1`:
![image](https://user-images.githubusercontent.com/8126674/31460697-e8ec63aa-aecf-11e7-8ce8-fe566fd2fb39.png)

ChShersh:
The coloring should be disabled for GHC >= 8.2.1. But also this should be added to `stack.yaml`:

```yaml
ghc-options:
  "$locals": -fhide-source-paths
```

ChShersh:
Okay, I propose to do the following with this `b` script:

1. It makes sense only for older GHC and for multi-package projects. I think nobody cares about pretty output for GHC < 8.2.1 because such old GHCs are the default nowadays for creating new project.
2. We should add "$locals" to every `stack.yaml` file.
3. We should remove coloring from `b` script.

PR #47, opened 2018-02-05
Windows compatibility

puhlenbruck:
Crazies like me like to work on windows sometimes.  This pr introduces a few changes to help improve the windows compatibility.

- Added a powershell install script and windows install instruction to README

- Replaced unicode chatacter → in prompt as it causes problems in
standard windows terminals

- created putStrFlush to be used instead of T.putStr, ensure content is
flushed to the console and prompts are displayed in the correct order

- replaced use of "rm" with System.Directory removeFile.  Wrapped in a
function deleteFile to continue on error

- modify setScriptCommand to do nothing on windows systems

*****

These changes should provide a basic level of support for windows platforms.  There are still some outstanding issues which are not addressed by this change:

- The build script feature still only produces a bash script.  It could potentially produce a powershell script as well for windows.

- The default background for the powershell console is blue,  the prompt arrow colour is also blue here making it a bit hard to read, perhaps the colour scheme could be adjusted.

- The tool still relies on curl to be installed for some functions.  Systems are not guaranteed to have this installed (this potentially affects linux systems as well).  For now I've added it to the prerequisite listing in the README.md, but maybe this download function could be done from the haskell code without requiring an external tool.

### DIFF
--- a/hs-init.hs
+++ b/hs-init.hs
@@ -431,7 +431,8 @@ falseMessage target = False <$ warningMessage (T.toTitle target <> " won't be ad
 -- Ansi-terminal
 ----------------------------------------------------------------------------
 
-putStrFlush :: Text -> IO()
+-- Explicit flush ensures prompt messages are in the correct order on all systems.
+putStrFlush :: Text -> IO ()
 putStrFlush msg = do
   T.putStr msg
   hFlush stdout
@@ -443,12 +444,14 @@ bold :: IO ()
 bold = setSGR [SetConsoleIntensity BoldIntensity]
 
 reset :: IO ()
-reset = setSGR [Reset]
+reset = do
+  setSGR [Reset]
+  hFlush stdout
 
 prompt :: IO Text
 prompt = do
   setColor Blue
-  putStrFlush " ->   "
+  putStrFlush "  ->   "
   reset
   T.getLine
 
@@ -516,9 +519,10 @@ instance (a ~ Text, b ~ ()) => IsString ([a] -> IO b) where
   fromString "cd" [arg] = setCurrentDirectory $ T.unpack arg
   fromString cmd args   = callProcess cmd (map T.unpack args)
 
+-- Delete file, but just print a message if delete fails and continue instead of raising an error.
 deleteFile :: FilePath -> IO  ()
 deleteFile file = catch (removeFile file) printError
-  where printError (e :: SomeException) = errorMessage $ "Could not delete file `" <> T.pack file <> "'. " <> T.pack  (displayException e)
+  where printError (e :: SomeException) = errorMessage $ "Could not delete file '" <> T.pack file <> "'. " <> T.pack  (displayException e)
 
 ----------------------------------------------------------------------------
 -- IO Questioning
--- a/hs-init.hs
+++ b/hs-init.hs
@@ -26,6 +26,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies        #-}
 
+import Control.Exception
 import Control.Monad (when)
 import Data.Aeson (FromJSON (..), decodeStrict, withObject, (.:))
 import Data.ByteString.Char8 (pack)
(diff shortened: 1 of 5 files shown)


vrom911:
Wow! @puhlenbruck thanks for contributing to `hs-init` :tada:
I haven't looked through the code yet, but I hope to do it as soon as possible.. Anyway, the idea is great, I would like to see this support in the project!

puhlenbruck:
@vrom911 Thanks! Please do let me know if there are any suggestions or changes you would like to see before merging.

hs-init.hs
```diff
@@ -508,7 +514,11 @@ customizeLicense l t nm year
 -- This is needed to be able to call commands by writing strings.
 instance (a ~ Text, b ~ ()) => IsString ([a] -> IO b) where
   fromString "cd" [arg] = setCurrentDirectory $ T.unpack arg
-  fromString cmd args   = callCommand $ showCommandForUser cmd (map T.unpack args)
+  fromString cmd args   = callProcess cmd (map T.unpack args)
+
+deleteFile :: FilePath -> IO  ()
+deleteFile file = catch (removeFile file) printError
+  where printError (e :: SomeException) = errorMessage $ "Could not delete file `" <> T.pack file <> "'. " <> T.pack  (displayException e)
```
vrom911: I see that  back-ticks are mismatching in here.

hs-init.hs
```diff
@@ -442,14 +448,14 @@ reset = setSGR [Reset]
 prompt :: IO Text
 prompt = do
   setColor Blue
-  T.putStr "  →   "
+  putStrFlush " ->   "
```
vrom911: Can I ask you to make here 2 spaces before the arrow, I think it adds a little bit more readability :slightly_smiling_face:

hs-init.hs
```diff
@@ -508,7 +514,11 @@ customizeLicense l t nm year
 -- This is needed to be able to call commands by writing strings.
 instance (a ~ Text, b ~ ()) => IsString ([a] -> IO b) where
   fromString "cd" [arg] = setCurrentDirectory $ T.unpack arg
-  fromString cmd args   = callCommand $ showCommandForUser cmd (map T.unpack args)
+  fromString cmd args   = callProcess cmd (map T.unpack args)
+
+deleteFile :: FilePath -> IO  ()
+deleteFile file = catch (removeFile file) printError
```
vrom911: I like this solution! Great! I think we can continue and this trash file that didn't removed is not the reason to stop the process

hs-init.hs
```diff
@@ -430,6 +431,11 @@ falseMessage target = False <$ warningMessage (T.toTitle target <> " won't be ad
 -- Ansi-terminal
 ----------------------------------------------------------------------------
 
+putStrFlush :: Text -> IO()
```
ChShersh: I think you forgot to add space before `(`

hs-init.hs
```diff
@@ -508,7 +514,11 @@ customizeLicense l t nm year
 -- This is needed to be able to call commands by writing strings.
 instance (a ~ Text, b ~ ()) => IsString ([a] -> IO b) where
   fromString "cd" [arg] = setCurrentDirectory $ T.unpack arg
-  fromString cmd args   = callCommand $ showCommandForUser cmd (map T.unpack args)
+  fromString cmd args   = callProcess cmd (map T.unpack args)
```
ChShersh: Could you please clarify the reason for changing `callCommand showCommandForUser` with `callProcess`?

ChShersh:
@puhlenbruck Regarding other issues you shared:

> The build script feature still only produces a bash script.

Well, I guess we don't need to create `b` script for GHC-8.2.2+, because script doesn't work for this GHC. And this version of GHC already has coloring and can strip folders from output. See relevant issue #27.

ChShersh:
Ahh, also about this PR: I think you should also add information to changelog regarding your changes 👍 Because your changes are great 🙂

hs-init.hs
puhlenbruck: Despite the docs claiming otherwise, `showCommandForUser` doesn't actually produce a string that works on windows.  The problem seems similar to https://github.com/haskell/process/issues/51

hs-init.hs
puhlenbruck: In case there is confusion, this function does allow the process to continue simply printing a message if the delete fails, but not stopping the process.

vrom911: Yes, exactly what is needed! :+1:

vrom911:
@puhlenbruck Thanks again for your wonderful contribution! This makes me so happy that `hs-init` becomes better! :heart:

Issue #3, opened 2017-08-19
Improve Travis settings created by 'summoner'


vrom911:
Implement matrix for different ghc versions support

vrom911:
Resolved in #49

PR #49, opened 2018-03-15
[#3] Improve Travis config, refactor

vrom911:
I have tested current settings for test projects. Build passes. But I had an issue with having `cabal >=2.0`. `setup --upgrade-cabal` failed with strange error on `8.0.2`..

### DIFF
--- a/hs-init.hs
+++ b/hs-init.hs
@@ -70,7 +70,7 @@ defaultLicense :: Text
 defaultLicense = "MIT"
 
 defaultGHC :: Text
-defaultGHC = "8.2.2"
+defaultGHC = ""
 
 currentYear :: IO Text
 currentYear = do
@@ -161,8 +161,9 @@ generateProject repo owner description Targets{..} = do
   bench  <- decisionToBool isBenchmark "benchmarks"
 
   putStrLn "Latest GHCs: 7.10.3 8.0.2 8.2.2"
+  putStrLn "The project will be created with the latest resolver for GHC-8.2.2."
   testedVersions <- T.words <$>
-    queryDef "Versions of GHC to test with (space-separated): " defaultGHC
+    queryDef "Additionally you can specify versions of GHC to test with (space-separated): " defaultGHC
   -- create stack project
   doStackCommands ProjectData{..}
   -- make b executable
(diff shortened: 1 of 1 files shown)


hs-init.hs
```diff
@@ -70,7 +70,7 @@ defaultLicense :: Text
 defaultLicense = "MIT"
 
 defaultGHC :: Text
-defaultGHC = "8.2.2"
+defaultGHC = ""
```
ChShersh: Maybe it's better to rename this variable as `defaultGHCs` since it represents multiple additional GHC versions to test with?

hs-init.hs
```diff
@@ -627,15 +630,15 @@ createStackTemplate
...
+    extra-doc-files:     README.md
+    cabal-version:       >=1.24
     $testedWith
     $endLine
     |]
 
   testedWith :: Text
-  testedWith = "tested-with:         " <>
-          T.intercalate ", " (map ("GHC == " <>) testedVersions)
+  testedWith = "tested-with:         GHC == 8.2.2" <>
```
ChShersh: I think we should have `defaultGHC` variable with value `8.2.2` and use it here. When we need to bump up to `GHC-8.4.1` we just change single variable instead of remembering what we need to change.

hs-init.hs
```diff
@@ -890,39 +886,82 @@ createStackTemplate
...
 
+  -- create specified @stack.yaml@ files
+  createStackYamls :: [Text] -> Text
+  createStackYamls = T.concat . map createStackYaml
+   where
+    createStackYaml :: Text -> Text
+    createStackYaml ghc =
+      let lts = case ghc of
+            "8.0.2"  -> "9.21"
+            "7.10.3" -> "6.35"
```
ChShersh: Unfortunately, we build without `-Wall` so GHC don't warn you that pattern matching is not exhaustive here :(

I guess it's okay for now. But in some future we should have something like:

```haskell
data GhcVer = Ghc7103 | Ghc802 | Ghc822 ...

parseGhcVer :: Text -> Maybe GhcVer
```

hs-init.hs
vrom911: I guess it's better just to remove it and put `""` right in the place where it's used as I'm sure we won't change his behavior. How about this?

Issue #48, opened 2018-02-07
Consider using 'stack install' for installing 'summoner`


ChShersh:
This was requested by many people. And it's probably a better idea for security reasons.

Basically this issue covers several things:

- [x] Convert script to project
- [x] Use some `.toml` file for configuration
- [x] Patch CLI interface and add `--version` command
- [x] Upload to Hackage

ChShersh:
Finally released!

PR #52, opened 2018-05-20
[#48] Convert script to project, rename to Summoner

vrom911:


src/Summoner/CLI.hs
```diff
@@ -0,0 +1,181 @@
...
+summon = execParser prsr >>= runWithOptions
+
+-- | Run 'hs-init' with cli options
+runWithOptions :: InitOpts -> IO ()
+runWithOptions (InitOpts projectName targets) = do
+     -- Generate the project.
+    generateProject projectName targets
+
+    bold
+    T.putStrLn "\nJob's done"
```
ChShersh: This looks similar to `boldDefault` function. Maybe you can create separate helper function for this to slightly remove code boilerplate?

src/Summoner/CLI.hs
```diff
@@ -0,0 +1,181 @@
...
+       , command "off" $ info (helper <*> targetsP Nop) (progDesc "Specify options to disable")
+       ]
+
+optsP :: Parser InitOpts
+optsP = do
+    projectName <- T.pack <$> strArgument (metavar "PROJECT_NAME")
+    on  <- optional onP
+    off <- optional offP
+
+    pure $ InitOpts projectName (fold $ on `mappend` off)
```
ChShersh: I think you can replace `mappend` with `<>` here

src/Summoner/CLI.hs
```diff
@@ -0,0 +1,181 @@
...
+
+offP :: Parser Targets
+offP = subparser $ mconcat
+       [ metavar "off [OPTIONS]"
+       , command "off" $ info (helper <*> targetsP Nop) (progDesc "Specify options to disable")
+       ]
+
+optsP :: Parser InitOpts
+optsP = do
+    projectName <- T.pack <$> strArgument (metavar "PROJECT_NAME")
```
ChShersh: I think you don't need `T.pack` here anymore. Latest version of `optparse-applicative` has polymorphic `strArgument` if I recall correctly.

src/Summoner/CLI.hs
```diff
@@ -0,0 +1,181 @@
...
+$endLine
+                                                  ___
+                                                /  .  \
+                                               │\_/│   │
+                                               │   │  /│
+  __________________________________________________-' │
+ ╱                                                     │
+╱   .-.                                                │
+│  /   \                                               │
+│ |\_.  │ hs-init — tool for creating Haskell projects │
```
ChShersh: Need to replace `hs-init` here with `summoner`... Unfortunately `summoner` is one letter longer... :disappointed:

src/Summoner/License.hs
```diff
@@ -0,0 +1,50 @@
...
+    , ("GPL-2",      "gpl-2.0")
+    , ("GPL-3",      "gpl-3.0")
+    , ("LGPL-2.1",   "lgpl-2.1")
+    , ("LGPL-3",     "lgpl-3.0")
+    , ("AGPL-3",     "agpl-3.0")
+    , ("Apache-2.0", "apache-2.0")
+    , ("MPL-2.0",    "mpl-2.0")
+    ]
+
+newtype License = License { lcnsText :: Text }
```
ChShersh: Maybe you could use `License` newtype in `githubLicense*` list and in `customizeLicense` function? With `IsString` instance this should be easier. But I guess this can improve code type safety.

src/Summoner/Project.hs
```diff
@@ -0,0 +1,196 @@
...
+import qualified Data.Text.IO as T
+
+-- | Used for detecting the user decision during CLI input.
+data Decision = Yes | Nop | Idk
+
+instance Semigroup Decision where
+    (<>) :: Decision -> Decision -> Decision
+    Idk <> x   = x
+    x   <> Idk = x
+    _   <> x   = x
```
ChShersh: I think that first test that can be added is semigroup and monoid laws for `Decision`.

src/Summoner/Project.hs
```diff
@@ -0,0 +1,196 @@
...
+
+          * Application                * Development
+          * Compilers/Interpreters     * Testing
+          * Web
+          * Game
+          * Utility
+
+        |]
+    category <- query "Category: "
+    license  <- choose "License: " $ nub (defaultLicense : licenseNames)
```
ChShersh: What do you think about adding `universum` to dependencies for `summoner`?

src/Summoner/Template.hs
```diff
@@ -0,0 +1,528 @@
...
+        bug-reports:         https://github.com/${owner}/${repo}/issues
+        license:             $license
+        license-file:        LICENSE
+        author:              $nm
+        maintainer:          $email
+        copyright:           $year $nm
+        category:            $category
+        build-type:          Simple
+        extra-doc-files:     README.md
+        cabal-version:       >=1.24
```
ChShersh: When I tested projects recently I noticed that `>=` shouldn't exist here. Just `1.24`.

src/Summoner/Template.hs
```diff
@@ -0,0 +1,528 @@
...
+        main :: IO ()
+        main = someFunc
+        $endLine
+        |]
+
+    createBenchmark :: Text
+    createBenchmark =
+      [text|
+      {-# START_FILE benchmark/Main.hs #-}
+      import Criterion.Main
```
ChShersh: Maybe we should use `gauge` as default benchmarking library?

src/Summoner/Template.hs
```diff
@@ -0,0 +1,528 @@
...
+        *.hp
+        *.eventlog
+        .virtualenv
+        .hsenv
+        .hpc
+        .cabal-sandbox/
+        cabal.sandbox.config
+        cabal.config
+        cabal.project.local
+        .HTF/
```
ChShersh: I think you also could add `.ghc.environment.*` to this list. Sometimes useful when you need to build project with `.cabal` as well. I needed it recently...

summoner.cabal
```diff
@@ -0,0 +1,68 @@
...
+bug-reports:         https://github.com/kowainik/summoner/issues
+license:             MPL-2.0
+license-file:        LICENSE
+author:              Kowainik
+maintainer:          xrom.xkov@gmail.com
+copyright:           2018 Kowainik
+category:            CLI, CLI Tool, Development
+build-type:          Simple
+extra-doc-files:     README.md
+cabal-version:       >=1.24
```
ChShersh: `>=` to remove here as well I think...

summoner.cabal
```diff
@@ -0,0 +1,68 @@
...
+  exposed-modules:     Summoner
+                         Summoner.Ansi
+                         Summoner.CLI
+                         Summoner.Default
+                         Summoner.License
+                         Summoner.Process
+                         Summoner.Project
+                         Summoner.ProjectData
+                         Summoner.Question
+                         Summoner.Template
```
ChShersh: I like your current module structure! Really helps to navigate between things :heart_eyes:

src/Summoner/Project.hs
vrom911: Hmm, that will increase dependencies a bit..

ChShersh: Yeah, that's right... When you have 123 vs. 124 dependency than having extra dependency is not really a problem if it increases productivity a lot! But since `summoner` is quite small let's wait for it to become bigger and see whether we can actually gain something from adding `universum`...

CHANGELOG.md
```diff
@@ -1,22 +1,32 @@
-1.0.4
+# Summoner
+
+1.0.0
+=====
+
+* __Important:__ Rename `hs-init` to `summoner`. Transform the script into the package.
```
ChShersh: 👍

PR #56, opened 2018-05-22
[#50] Introduce data type for GHC version

### LINKED ISSUE #50: Better GHC version parsing
Idea in short words:

```haskell
data GhcVer = Ghc7103 | Ghc802 | Ghc822 ...

parseGhcVer :: Text -> Maybe GhcVer
```

ChShersh:
Resolves #50

### DIFF
--- a/src/Summoner/Project.hs
+++ b/src/Summoner/Project.hs
@@ -13,6 +13,7 @@ import Control.Monad (when)
 import Data.Aeson (decodeStrict)
 import Data.ByteString.Char8 (pack)
 import Data.List (nub)
+import Data.Maybe (catMaybes)
 import Data.Semigroup (Semigroup (..))
 import Data.Text (Text)
 import NeatInterpolation (text)
@@ -24,7 +25,7 @@ import Summoner.Default (currentYear, defaultEmail, defaultGHC, defaultLicense, 
                          defaultOwner)
 import Summoner.License (License (..), customizeLicense, githubLicenseQueryNames, licenseNames)
 import Summoner.Process (deleteFile)
-import Summoner.ProjectData (ProjectData (..))
+import Summoner.ProjectData (ProjectData (..), parseGhcVer, showGhcVer)
 import Summoner.Question (checkUniqueName, choose, query, queryDef)
 import Summoner.Template (createStackTemplate)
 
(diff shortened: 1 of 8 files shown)


src/Summoner/Project.hs
```diff
@@ -153,10 +154,11 @@ generateProject projectName Targets{..} = do
     test   <- decisionToBool isTest "tests"
     bench  <- decisionToBool isBenchmark "benchmarks"
 
-    putStrLn "Latest GHCs: 7.10.3 8.0.2 8.2.2"
-    putStrLn $ "The project will be created with the latest resolver for GHC-" ++ T.unpack defaultGHC
-    testedVersions <- T.words <$>
+    T.putStrLn "Latest GHCs: 7.10.3 8.0.1 8.0.2 8.2.2"
```
ChShersh: Maybe here it's better to tell something like `Latest supported by summoner GHCs`? 🤔

vrom911: I don't really like this plain text, as we should be keeping in mind changing it when we change the list of supported versions..

src/Summoner/Project.hs
```diff
@@ -153,10 +154,11 @@ generateProject projectName Targets{..} = do
...
+    T.putStrLn $ "The project will be created with the latest resolver for GHC-" <> showGhcVer defaultGHC
```
vrom911: Here also, as you mentioned above, it's not the latest, but default (or latest for summoner, but can be changed by user).

src/Summoner/Project.hs
```diff
@@ -153,10 +154,11 @@ generateProject projectName Targets{..} = do
...
+    testedVersions <- catMaybes . map parseGhcVer . T.words <$>
       queryDef "Additionally you can specify versions of GHC to test with (space-separated): " ""
+    -- TODO: tell which ghc versions were failing to parse?...
```
vrom911: What do you mean?

src/Summoner/ProjectData.hs
```diff
@@ -23,5 +26,27 @@ data ProjectData = ProjectData
...
+showGhcVer Ghc802  = "8.0.2"
+showGhcVer Ghc822  = "8.2.2"
+
+-- | Converts numeric dot-separated GHC version into 'GhcVer'.
+parseGhcVer :: Text -> Maybe GhcVer
+parseGhcVer "7.10.3" = Just Ghc7103
+parseGhcVer "8.0.1"  = Just Ghc801
+parseGhcVer "8.0.2"  = Just Ghc802
+parseGhcVer "8.2.2"  = Just Ghc822
+parseGhcVer _        = Nothing
```
vrom911: Great work, thank you!

src/Summoner/Project.hs
ChShersh: @vrom911 Maybe we could have something like that `allGhcVersions = [minBound .. maxBound]` to not forget to update this list.

src/Summoner/Project.hs
ChShersh: `parseGhcVer` function returns `Maybe GhcVer`. This means that it can fail if user enters something like `8..0.2` or `8.2.3`. Probably some helpful message would be great like:

```haskell
Invalid specified GHC versions: '8..0.2' '8.2.3'
```

And probably if user enters something like `8.2.3` we shouldn't just ignore this input and use default GHC, maybe we should ask for retry...

vrom911: Yes, that's true, something should be done about that..
I guess that just ignoring is not the right thing to do here, as the input could be several versions, and typos are very likely to happen, so the particular error message should be shown and the question re-asked. Are you agree?

ChShersh: @vrom911 I think the same. Will to do something about that.

PR #59, opened 2018-05-23
[#48] Add configurations, read from the file

vrom911:
Things to do
- [x] Handle `[GhcVer]` from the custom configurations
- [x] Add new instructions to the `README` explaining how configs should work

### DIFF
--- a/src/Summoner.hs
+++ b/src/Summoner.hs
@@ -1,6 +1,7 @@
 module Summoner
        ( module Summoner.Ansi
        , module Summoner.CLI
+       , module Summoner.Config
        , module Summoner.Default
        , module Summoner.License
        , module Summoner.Process
@@ -12,6 +13,7 @@ module Summoner
 
 import Summoner.Ansi
 import Summoner.CLI
+import Summoner.Config
 import Summoner.Default
 import Summoner.License
 import Summoner.Process
--- a/src/Summoner.hs
+++ b/src/Summoner.hs
@@ -9,6 +9,7 @@ module Summoner
        , module Summoner.ProjectData
        , module Summoner.Question
        , module Summoner.Template
+       , module Summoner.Validation
        ) where
 
 import Summoner.Ansi
@@ -21,3 +22,4 @@ import Summoner.Project
 import Summoner.ProjectData
 import Summoner.Question
 import Summoner.Template
+import Summoner.Validation
--- a/src/Summoner/CLI.hs
+++ b/src/Summoner/CLI.hs
@@ -16,12 +16,14 @@ import Options.Applicative (Parser, ParserInfo, command, execParser, flag, fullD
                             strArgument, strOption, subparser)
 import Options.Applicative.Help.Chunk (stringChunk)
 import System.Directory (doesFileExist, getHomeDirectory)
+import System.FilePath ((</>))
 
 import Summoner.Ansi (boldText)
 import Summoner.Config (ConfigP (..), PartialConfig, defaultConfig, finalise, loadFileConfig)
 import Summoner.Default (endLine)
 import Summoner.Project (generateProject)
 import Summoner.ProjectData (Decision (..))
+import Summoner.Validation (Validation (..))
 
 import qualified Data.Text as T
 
@@ -36,30 +38,27 @@ summon = execParser prsr >>= runWithOptions
 runWithOptions :: InitOpts -> IO ()
 runWithOptions (InitOpts projectName cliConfig maybeFile) = do
     file <- case maybeFile of
-        Nothing -> getHomeDirectory >>= pure . (++ "/summoner.toml")
+        Nothing -> (</> "summoner.toml") <$> getHomeDirectory
         Just x  -> pure x
     isFile <- doesFileExist file
-    config <-
-        if isFile then do
-            parsedFile <- loadFileConfig file
-            case parsedFile of
-                Left msg -> do
-                    putStrLn $ "Couldn't parse file " ++ file ++ ": " ++ show msg
-                    putStrLn "Using default config.."
-                    pure defaultConfig
-                Right conf -> pure $ defaultConfig <> conf
-        else pure defaultConfig
-     -- get the final config
-    let globalConfig = case finalise (config <> cliConfig) of
-             Left msg -> error $ T.unpack msg
-             Right c  -> c
+    fileConfig <-
+        if isFile
+        then loadFileConfig file
+        else pure mempty
+    -- union all possible configs
+    let unionConfig = defaultConfig <> fileConfig <> cliConfig
+    -- get the final config
+    let finalConfig = case finalise unionConfig of
+             Failure msgs -> error $ T.unpack $ T.intercalate "\n" msgs
+             Success c    -> c
     -- Generate the project.
-    generateProject projectName globalConfig
+    generateProject projectName finalConfig
 
     boldText "\nJob's done\n"
 
 -- | Initial parsed options from cli
-data InitOpts = InitOpts Text PartialConfig (Maybe FilePath)   -- ^ Includes the project name and target options.
+data InitOpts = InitOpts Text PartialConfig (Maybe FilePath)
+    -- ^ Includes the project name, config from the CLI and possible file where custom congifs are.
 
 targetsP ::  Decision -> Parser PartialConfig
 targetsP d = do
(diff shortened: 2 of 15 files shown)


src/Summoner/Config.hs
```diff
@@ -0,0 +1,170 @@
...
+    <*> pure cAppVey
+    <*> pure cPrivate
+    <*> pure cScript
+    <*> pure cLib
+    <*> pure cExe
+    <*> pure cTest
+    <*> pure cBench
+    <*> fin  "ghcVersions" cGhcVer
+  where
+    fin name = maybe (Left $ "Missing field: " <> name) Right . getLast
```
ChShersh: Actually, this is an interesting piece of code. `Applicative` instance for `Either` works in that way that it fails on first error and returns only single error to user. But it would be much more convenient to tell all absent fields at once. This actually already exist in Haskell world. There's data type called [`Validation`](https://hackage.haskell.org/package/either-5/docs/Data-Either-Validation.html#t:Validation). It's like `Either` (just different constructor names). But it has slightly more interesting [`applicative instance`](https://hackage.haskell.org/package/either-5/docs/src/Data.Either.Validation.html#line-57). It accumulates all errors. So we can copy-paste `Validation` to us and this instance and just use it here instead of `Either` to output multiple errors to user regarding absent fields!

src/Summoner/Config.hs
```diff
@@ -0,0 +1,170 @@
...
+
+-- | Read configuration from the given file and return it in data type.
+loadFileConfig :: MonadIO m => FilePath -> m (Either Toml.EncodeException PartialConfig)
+loadFileConfig filePath = liftIO $ TIO.readFile filePath >>= pure . Toml.encode configT
```
ChShersh: I think you can just use `throwIO` here instead of returning `Either`. If there's error in configuration we don't want to skip this file, we want to throw exception. This also will simplify code a bit in `main`.

src/Summoner/Project.hs
```diff
@@ -198,3 +133,23 @@ generateProject projectName Targets{..} = do
         "git" ["add", "."]
         "git" ["commit", "-m", "Create the project"]
         "git" ["push", "-u", "origin", "master"]
+
+    categoryText :: Text
```
ChShersh: Cool idea to move this under `where`. 
You get the joke?) under `where` is like _underwear_ :joy:

src/Summoner/ProjectData.hs
```diff
@@ -1,12 +1,22 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric  #-}
+{-# LANGUAGE InstanceSigs   #-}
```
ChShersh: I think `DeriveGeneric` and `InstanceSigs` should be put into default extensions safely.

src/Summoner/CLI.hs
```diff
@@ -31,27 +34,55 @@ summon = execParser prsr >>= runWithOptions
 
 -- | Run 'hs-init' with cli options
 runWithOptions :: InitOpts -> IO ()
-runWithOptions (InitOpts projectName targets) = do
-     -- Generate the project.
-    generateProject projectName targets
+runWithOptions (InitOpts projectName cliConfig maybeFile) = do
+    file <- case maybeFile of
```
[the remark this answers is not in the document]
vrom911: I'm agree with adding it after so many stupid imports..

src/Summoner.hs
```diff
@@ -1,6 +1,7 @@
 module Summoner
        ( module Summoner.Ansi
        , module Summoner.CLI
+       , module Summoner.Config
        , module Summoner.Default
```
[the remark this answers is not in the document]
vrom911: Yes, it should be

src/Summoner/Config.hs
```diff
@@ -0,0 +1,170 @@
...
+{-# LANGUAGE DeriveAnyClass       #-}
+{-# LANGUAGE DeriveGeneric        #-}
+{-# LANGUAGE FlexibleContexts     #-}
+{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE KindSignatures       #-}
+{-# LANGUAGE StandaloneDeriving   #-}
+{-# LANGUAGE TypeFamilies         #-}
+{-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
```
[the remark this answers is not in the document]
vrom911: 😆

src/Summoner/CLI.hs
```diff
@@ -31,27 +36,52 @@ summon = execParser prsr >>= runWithOptions
...
+        Nothing -> (</> "summoner.toml") <$> getHomeDirectory
```
ChShersh: Maybe we can put `"summoner.toml"` into `Default` module. As well as `IO` function which returns full path to this file.

src/Summoner/CLI.hs
```diff
@@ -31,27 +36,52 @@ summon = execParser prsr >>= runWithOptions
...
+        Just x  -> pure x
+    isFile <- doesFileExist file
+    fileConfig <-
+        if isFile
+        then loadFileConfig file
+        else pure mempty
```
ChShersh: After thinking a little bit more on this, I came to decision that logic here should be slightly complicated to increase UX.

1. If file exist, print information message (probably we can use `Blue` for information?) that this file configuration will be used.
2. If this is a default file `~/summoner.toml` and it doesn't exist then print info message that no configuration file was specified (or will be used).
3. If this is not a default configuration file (i.e. specified by `--file`) then we should abort immediately because I think that if you specify file manually through CLI you expect this file to exist. But if you have typo in file name, it's better to see error that such file doesn't exist and fix CLI command instead of silently continue to work.

src/Summoner/CLI.hs
```diff
@@ -37,20 +38,30 @@ summon = execParser prsr >>= runWithOptions
...
+        Just x  -> pure (False, x)
     isFile <- doesFileExist file
     fileConfig <-
         if isFile
-        then loadFileConfig file
-        else pure mempty
+        then do
+            infoMessage $ "Configurations from " <> T.pack file <> " will be used."
+            loadFileConfig file
+        else  if isDefault
```
ChShersh: Extra space here..

src/Summoner/CLI.hs
```diff
@@ -37,20 +38,30 @@ summon = execParser prsr >>= runWithOptions
...
+                  exitFailure
     -- union all possible configs
     let unionConfig = defaultConfig <> fileConfig <> cliConfig
     -- get the final config
-    let finalConfig = case finalise unionConfig of
-             Failure msgs -> error $ T.unpack $ T.intercalate "\n" msgs
-             Success c    -> c
+    finalConfig <- case finalise unionConfig of
+             Failure msgs -> do
+                 for_ msgs $ \msg -> errorMessage msg
```
ChShersh: You can eta-reduce lambda here

src/Summoner/CLI.hs
```diff
@@ -37,20 +38,30 @@ summon = execParser prsr >>= runWithOptions
...
     fileConfig <-
         if isFile
-        then loadFileConfig file
-        else pure mempty
+        then do
+            infoMessage $ "Configurations from " <> T.pack file <> " will be used."
+            loadFileConfig file
+        else  if isDefault
+              then do
+                  warningMessage "Default config file is missing."
```
ChShersh: I think you can add path `defaultTomlFile` path to this message just in case. So stupid people won't complain in a way like _It says to me that default config file is missing but I don't know where default config file should be_.

src/Summoner/Question.hs
```diff
@@ -63,9 +63,8 @@ queryDef question defAnswer = do
     if | T.null answer -> pure defAnswer
        | otherwise     -> pure answer
 
-queryManyRepeatOnFail :: forall a . (Text -> Maybe a) -> Text -> IO [a]
-queryManyRepeatOnFail parser question = do
-    T.putStrLn question
+queryManyRepeatOnFail :: forall a . (Text -> Maybe a) -> IO [a]
+queryManyRepeatOnFail parser = do
```
ChShersh: You don't need `do` here and you can put `promptLoop` on the same line as `queryManyRepeatOnFail`

PR #64, opened 2018-05-26
[#60] Use universum as default prelude

### LINKED ISSUE #60: Move to `universum`
Really, it's time

ChShersh:
Resolves #60

### DIFF
--- a/src/Prelude.hs
+++ b/src/Prelude.hs
@@ -1,0 +1,16 @@
+-- | Uses [universum](http://hackage.haskell.org/package/universum) as default prelude.
+
+module Prelude
+       ( module Universum
+       , intercalateMap
+       ) where
+
+import Universum hiding (Key)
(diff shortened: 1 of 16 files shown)


src/Prelude.hs
```diff
@@ -0,0 +1,16 @@
...
+
+import Universum hiding (Key)
+
+import qualified Data.Text as T
+
+-- | Converts every element of list into 'Text' and then joins every element
+-- into single 'Text' like 'T.intercalate'.
+intercalateMap :: Text -> (a -> Text) -> [a] -> Text
+intercalateMap between showT = T.intercalate between . map showT
+-- TODO: maybe Prelude is not the best place for such function?...
```
vrom911: Hmm, probably it'snot, because it's hard to determine where this unusual function comes from..

PR #68, opened 2018-05-26
[#38] Add option for default extensions in config

### LINKED ISSUE #38: Add some default-extensions
Often people want to have some `default-extensions` inside their `.cabal` file. I'm not sure how interface should look like because not everyone likes `default-extensions` and not eveyryone agrees on which extensions should be enabled by default.

vrom911:
Resolves #38

### DIFF
--- a/src/Summoner/Config.hs
+++ b/src/Summoner/Config.hs
@@ -60,6 +60,7 @@ data ConfigP (p :: Phase) = Config
     , cBench          :: Decision
     , cPreludePackage :: Last Text
     , cPreludeModule  :: Last Text
+    , cExtensions     :: [Text]
     } deriving (Generic)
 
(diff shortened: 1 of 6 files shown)


src/Summoner/Config.hs
```diff
@@ -136,6 +139,9 @@ configT = Config
     license :: Key -> BiToml License
     license =  dimapBijection unLicense License . Toml.str
 
+    extensions :: Key -> BiToml [Text]
+    extensions = dimapBijection Just (fromMaybe []) . Toml.maybeP (Toml.arrayOf Toml.strV)
```
ChShersh: `maybeToMonoid` is one character lower than `(fromMaybe [])` :slightly_smiling_face:

PR #72, opened 2018-05-27
[#67] Add Answer data type

### LINKED ISSUE #67: Support "Y" and "yes" and similar, not only 'y' and 'n' as answers
Basically we should create data type like this:

```haskell
data Answer = Yes | No
```

And then write custom parser for this type.

vrom911:
Resolves #67

### DIFF
--- a/src/Summoner/Project.hs
+++ b/src/Summoner/Project.hs
@@ -12,53 +12,25 @@ import NeatInterpolation (text)
 import System.Info (os)
 import System.Process (readProcess)
 
-import Summoner.Ansi (Color (..), beautyPrint, bold, infoMessage, italic, setColor, skipMessage,
-                      successMessage, warningMessage)
+import Summoner.Ansi (infoMessage, successMessage, warningMessage)
 import Summoner.Config (Config, ConfigP (..))
 import Summoner.Default (currentYear, defaultGHC)
 import Summoner.License (License (..), customizeLicense, githubLicenseQueryNames, licenseNames)
 import Summoner.Process ()
 import Summoner.ProjectData (CustomPrelude (..), Decision (..), ProjectData (..), parseGhcVer,
                              showGhcVer, supportedGhcVers)
-import Summoner.Question (checkUniqueName, choose, query, queryDef, queryManyRepeatOnFail)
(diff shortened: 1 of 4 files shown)


src/Summoner/Project.hs
```diff
@@ -12,36 +12,26 @@ import NeatInterpolation (text)
...
-            "y" -> trueMessage  target
-            "n" -> falseMessage target
-            _   -> error "Impossible happened"
-
-trueMessage, falseMessage :: Text -> IO Bool
-trueMessage  target = True  <$ successMessage (T.toTitle target <> " will be added to the project")
-falseMessage target = False <$ skipMessage (T.toTitle target <> " won't be added to the project")
-
+    Yes -> True  <$ trueMessage  target
+    Nop -> False <$ falseMessage target
```
ChShersh: I see that you do `True <$` and `False <$` for every use site of `trueMessage` and `falseMessage`. Earlier `trueMessage` was written like `True <$ successMessage ... smth ...`. Maybe I'm missing something. But is there reason to make return types of `trueMessage` and `falseMessage` be `IO ()` instead of previous `IO Bool`?

src/Summoner/Question.hs
```diff
@@ -43,6 +50,27 @@ choose question choices = do
            errorMessage "This wasn't a valid choice."
            choose question choices
 
+chooseYesNo :: Text -- ^ target
+            -> IO a -- ^ action for 'Y' answer
+            -> IO a -- ^ action for 'N' answer
```
ChShersh: Nice solution with using `IO` actions as arguments for this function! Really high-level interface :+1:

src/Summoner/Project.hs
vrom911: The reason is that I'm using it in the `chooseYesNo` where return type is `IO a` not `IO Bool`.. But probably I just need to write separate functions for messages itself..

ChShersh: @vrom911 I see. But in `chooseYesNo` function you wrote like this:

```haskell
trueMessage target >> yesDo
```

Since you use `>>` it doesn't matter what is the return type of `trueMessage`. Probably you wrote just `true` and `falseMessage` earlier and saw type errors, but after you refactored, compiler didn't warn you that you can return types back. I think it should work if you make type as `IO Bool`.

PR #79, opened 2018-05-28
[#74] Add showTree function

### LINKED ISSUE #74: Print hierarchy tree on project creating

vrom911:
Resolves #74

### DIFF
--- a/src/Summoner/Tree.hs
+++ b/src/Summoner/Tree.hs
@@ -1,6 +1,9 @@
+{-# LANGUAGE ViewPatterns #-}
+
 module Summoner.Tree
        ( TreeFs (..)
        , traverseTree
+       , showTree
        ) where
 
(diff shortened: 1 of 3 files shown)


src/Summoner/Tree.hs
```diff
@@ -19,3 +22,28 @@ traverseTree (Dir name children) = do
     createDirectoryIfMissing False name
     withCurrentDirectory name $ for_ children traverseTree
 traverseTree (File name content) = writeFile name content
+
+-- | Pretty shows the directory tree content.
+showTree :: TreeFs -> Text
+showTree = unlines . showOne "  " "" ""
```
ChShersh: Nice implementation! Probably we can do better in some future (to make output look like output of `tree` utilitly) but it requires more thinking.. And it's not important now.

PR #81, opened 2018-05-30
UX improvements

vrom911:


src/Summoner/CLI.hs
```diff
@@ -247,11 +247,24 @@ $endLine
...
+ `╲.  ╱╱  (   │  ║
+  ╲.╲╱        │  ║
+  `╰══════════╯
+
+
+
+
+
+
+
```
ChShersh: Maybe these empty lines are redundant and can be replaced with single `$endLine`?

src/Summoner/Template.hs
```diff
@@ -12,6 +12,8 @@ import NeatInterpolation (text)
 import Summoner.Default (defaultGHC, endLine)
 import Summoner.ProjectData (CustomPrelude (..), GhcVer (..), ProjectData (..), baseNopreludeVer,
                              latestLts, showGhcVer)
+import Summoner.Text (packageToModule)
+import Summoner.Text (intercalateMap)
```
ChShersh: These two lines can be joined into single line

src/Summoner/Text.hs
```diff
@@ -0,0 +1,16 @@
...
+       ( packageToModule
+       , intercalateMap
+       ) where
+
+import qualified Data.Text as T
+
+-- | Creates module name from the name of the package
+-- Ex: @my-lovely-project@ — @MyLovelyProject@
+packageToModule :: Text -> Text
+packageToModule = T.concat . map T.toTitle . T.splitOn "-"
```
ChShersh: You can move `headToUpper` function to this module as well and use it here instead of `toTitle`.

PR #93, opened 2018-06-11
[#92] Remove -fhide-source-paths for GHC < 8.2

### LINKED ISSUE #92: Travis fails for GHC < 8.2
As title, because `-fhide-source-paths` was only added in GHC 8.2.  Maybe only add it conditionally to the template?

adituv:
Fixes #92

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Summoner
 
+1.0.3
+=====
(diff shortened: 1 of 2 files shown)


ChShersh:
Just uploaded `summoner-1.0.3` to Hackage:

* https://hackage.haskell.org/package/summoner-1.0.3

PR #95, opened 2018-06-12
Update contributing guide

ChShersh:


CONTRIBUTING.md
```diff
@@ -11,8 +11,9 @@ If you found any bugs or have any proposals how to make this project better
 don't hesitate to create issues [here][issues] in a free format.
 
 #### Create a PR
-We love pull requests from everyone :heart:.
-To get started with this you should first fork, then clone the repo:
+We love pull requests from everyone :heart:. But, please, tell about your intention to work on something under corresponding issue.
```
vrom911: You can say about creating one if there is no such an issue

vrom911: Also below there is a note about commit message. `if` statement should be removed from there as we want the contribution only on existing issues (if my understanding is correct)

CONTRIBUTING.md
```diff
@@ -40,4 +40,4 @@ Thanks for spending time on reading this contributing guide! :sparkling_heart:
 
 [pr]: https://github.com/kowainik/summoner/compare/
 [issues]: https://github.com/kowainik/summoner/issues/new
-[stylish]:
+[stylish]: https://github.com/kowainik/summoner/blob/master/.stylish-haskell.yaml
```
vrom911: Thanks!

PR #98, opened 2018-06-28
Bump up to tomland-0.3

ChShersh:
Changes version of `tomland` to `0.3` and adds upper-bound to it.

I've tested it on our `.summoner.toml` config, it works.


stack-8.4.3.yaml
```diff
@@ -0,0 +1,4 @@
+resolver: nightly-2018-06-27
```
vrom911: Can you please also add Travis check for this as well?

PR #103, opened 2018-07-15
Add CI badges depending on user choise

vrom911:


src/Summoner/Template.hs
```diff
@@ -267,14 +267,24 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
         hackageLink :: Text =
           "https://hackage.haskell.org/package/" <> repo
+
         travisShield :: Text =
           "https://secure.travis-ci.org/" <> owner <> "/" <> repo <> ".svg"
         travisLink :: Text =
           "https://travis-ci.org/" <> owner <> "/" <> repo
+        travisBadge :: Text = if travis
+            then [text|[![Build status](${travisShield})](${travisLink})|]
+            else ""
```
ChShersh: I think you can use `emptyIfNot` function for such cases:

https://github.com/kowainik/summoner/blob/108f79c223046f2b8b8591b3e6e16cf95eada736/src/Summoner/Template.hs#L27-L28

PR #104, opened 2018-07-15
[#97] Add travis for Cabal

### LINKED ISSUE #97: Travis for `cabal`
This is subtask of #96 , but I would like to have a place for discussion of Travis settings.

The way to do `.travis.yaml` for cabal project that I know is [`haskell-ci`](https://github.com/haskell-CI/haskell-ci) (I used this solution for `o-clock` when it started, but then it was replaced with much simpler one). So I'm not sure which way to go here as I don't know how the support of several `ghc`s with `cabal` at the same time will affect our settings (if not to choose `haskell-ci`). And also what to do for both `cabal` and `stack` option.

### LINKED ISSUE #99: Use Travis-specific env variable TRAVIS_BUILD_DIR
Now paths to `stack.yaml` files look like this:

```
STACK_YAML="$HOME/build/kowainik/summoner/stack-$GHCVER.yaml"
```

We can use travis env variables to make this shorter:

```
STACK_YAML="$TRAVIS_BUILD_DIR/stack-$GHCVER.yaml"
```

vrom911:
Resolves #97 
Resolves #99 

Tested on [this](https://github.com/vrom911/new) project. See generated Travis file: https://github.com/vrom911/new/blob/master/.travis.yml . Note, that `cabal new-test` fails if there are no tests, so I made another check and making simple `echo` if test stanza was not chosen.

### DIFF
--- a/src/Summoner/Template.hs
+++ b/src/Summoner/Template.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE QuasiQuotes         #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE ViewPatterns        #-}
 
 -- | This module contains functions for stack template creation.
 
@@ -369,8 +370,9 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
     -- create travis.yml template
     travisYml :: Text
     travisYml =
-        let travisMtr = T.concat (map (travisMatrixItem . showGhcVer) (delete defaultGHC testedVersions))
-            defGhc    = showGhcVer defaultGHC in
+        let travisStackMtr = T.concat $ map travisStackMatrixItem (delete defaultGHC testedVersions)
+            travisCabalMtr = T.concat $ map travisCabalMatrixItem testedVersions
+            defGhc = showGhcVer defaultGHC in
         [text|
         sudo: true
         language: haskell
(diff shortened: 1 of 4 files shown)


src/Summoner/Template.hs
```diff
@@ -369,7 +370,8 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
     -- create travis.yml template
     travisYml :: Text
     travisYml =
-        let travisMtr = T.concat (map (travisMatrixItem . showGhcVer) (delete defaultGHC testedVersions))
+        let travisStackMtr = T.concat (map (travisStackMatrixItem . showGhcVer) (delete defaultGHC testedVersions))
```
ChShersh: I think you can make this line more consistent with `cabal` matrix item and move `showGhcVer` as a `ViewPattern` inside `travisStackMatrixItem` function.

src/Summoner/Template.hs
```diff
@@ -381,53 +383,83 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
 
         notifications:
           email: false
         $endLine
         |]
 
-    travisMatrixItem :: Text -> Text
-    travisMatrixItem ghcV =
+    cabalTest :: Text
+    cabalTest = if test then "cabal new-test" else "echo 'No tests'"
```
ChShersh: Looks like it deserves issue in `cabal` (if it doesn't exist yet) :pepe:

vrom911: https://github.com/haskell/cabal/issues/5435

PR #111, opened 2018-07-30
[#82] Add stackage badges

### LINKED ISSUE #82: Add stackage badges to README
Just like here:

* https://github.com/kowainik/relude

willbasky:
Resolves #82

### DIFF
--- a/src/Summoner/Template.hs
+++ b/src/Summoner/Template.hs
@@ -270,21 +270,21 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
         hackageLink :: Text =
           "https://hackage.haskell.org/package/" <> repo
 
-        stackBadges :: Text = stackBadgeLts <> stackBadgeNightly
-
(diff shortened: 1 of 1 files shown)


src/Summoner/Template.hs
```diff
@@ -269,6 +270,22 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
         hackageLink :: Text =
           "https://hackage.haskell.org/package/" <> repo
 
+        stackBadges :: Text = stackBadgeLts <> stackBadgeNightly
```
ChShersh: I think that using `<>` here doesn't add desired line-break. Anyway, I think that some redundant code can be removed and you can just using single quasi-quoter for this variable:

```haskell
[text|
![Stackage Lts](${stackShieldLts})](${stackLinkLts})
![Stackage Nightly](${stackShieldNightly})](${stackLinkNightly})
|]
```

Issue #110, opened 2018-07-26
Bring `life-sync` feature to print executing command


ChShersh:
Change instance to this one:

```haskell
instance (a ~ Text, b ~ ()) => IsString ([a] -> IO b) where
    fromString cmd args = do
        let cmdStr = showCommandForUser cmd (map toString args)
        putStrLn $ "⚙  " ++ cmdStr
        callCommand cmdStr
```

willbasky:
What is this feature about? Where is piece of code from?

ChShersh:
@willbasky This issue is about changing this instance in `summoner`:

https://github.com/kowainik/summoner/blob/6b1c87e23c1c8f55f706c615869af8fa73cb6cd7/src/Summoner/Process.hs#L23-L26

to prettier instance from `life-sync`:

https://github.com/kowainik/life-sync/blob/e537e372e34539ad1325e2685687d61f68869eb4/src/Life/Shell.hs#L45-L50

PR #115, opened 2018-08-02
[#110] Print executing command

willbasky:
Resolves [#110](https://github.com/kowainik/summoner/issues/110)

### DIFF
--- a/src/Summoner/Process.hs
+++ b/src/Summoner/Process.hs
@@ -11,8 +11,8 @@ module Summoner.Process
 import Relude
 
 import Control.Exception (catch, displayException)
-import System.Directory (removeFile, setCurrentDirectory)
-import System.Process (callProcess)
+import System.Directory (removeFile)
+import System.Process (callCommand, showCommandForUser)
 
 import Summoner.Ansi (errorMessage)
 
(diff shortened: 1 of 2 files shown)


ChShersh:
@willbasky Also, comment message should be `Resolves` instead of `Resolving`. Unfortunately, GitHub doesn't recognise all words.

ChShersh:
Okay, I tested it locally. It works as expected!

src/Summoner/Process.hs
```diff
@@ -22,8 +22,10 @@ import Summoner.Ansi (errorMessage)
 
 -- This is needed to be able to call commands by writing strings.
 instance (a ~ Text, b ~ ()) => IsString ([a] -> IO b) where
-    fromString "cd" [arg] = setCurrentDirectory $ toString arg
-    fromString cmd args   = callProcess cmd (map toString args)
+    fromString cmd args = do
+        let cmdStr = showCommandForUser cmd (map toString args)
+        putStrLn $ "⚙  " ++ cmdStr
+        callCommand cmdStr
```
ChShersh: You should also update changelog 🙂

willbasky: ops! Sure..

Issue #91, opened 2018-06-08
Add 'warnings' field to configuration


chshersh:
There're some warnings which are not included in `-Wall` but some people think they should be used anyway. I think we should create special configuration field for this. And add some fields to our default config.

Source:

* https://www.reddit.com/r/haskell/comments/6q9tcp/ghc_warnings_you_should_use_in_addition_to_wall/

Comments are interesting as well.

ChShersh:
@willbasky This task is about only adding `warnings` fields to configuration. You should this extra field to configuration here:

https://github.com/kowainik/summoner/blob/96b10338c4296c2363913f15236ea49a602b14f0/src/Summoner/Config.hs#L45-L65

And then you should use it to populate `ghc-options` field in `.cabal` files.

willbasky:
Will warnings add to configuration if there is `--file conf.toml` or\and if _yes_ in interactive mode?

ChShersh:
@willbasky We have this note in README about priority order:

* https://github.com/kowainik/summoner#usage

But personally I think that it doesn't make much sense to ask about warnings interactively. We already add `-Wall` by default and this should be enough in 99.9% cases. For those who want extra warnings, it's better to specify them in configuration because names on warnings are very long. And it's not convenient to type them manually. So I expect that nobody will write warnings in interactive mode.

But I think that it's a good idea to be able to specify them through CLI options.

willbasky:
I agree with you. If someone will want to change warnings it easy to rewrite them in text mode after project created.

vrom911:
I don't entirely agree about CLI part. I guess it's almost the same as with pragmas, better to use config files for that.

ChShersh:
@vrom911 Yeah, that makes sense as well. Probably it's better to not allow them as CLI as well.

PR #116, opened 2018-08-03
[#91] Add warning fields

willbasky:
Resolves [#91](https://github.com/kowainik/summoner/issues/91)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
   Fix travis cache directories
 * [#110](https://github.com/kowainik/summoner/issues/110):
   Print executing command
+* [#91](https://github.com/kowainik/summoner/issues/91):
+  Add warning fields for ghc
 
 1.0.6
 =====
--- a/src/Summoner/Config.hs
+++ b/src/Summoner/Config.hs
@@ -62,6 +62,7 @@ data ConfigP (p :: Phase) = Config
     , cBench      :: Decision
     , cPrelude    :: Last CustomPrelude
     , cExtensions :: [Text]
+    , cWarnings   :: [Text]
     } deriving (Generic)
 
 deriving instance (GSemigroup (p :- Text), GSemigroup (p :- License), GSemigroup (p :- [GhcVer])) => GSemigroup (ConfigP p)
@@ -106,6 +107,7 @@ defaultConfig = Config
     , cBench    = Idk
     , cPrelude  = Last Nothing
     , cExtensions = []
+    , cWarnings = []
     }
 
 -- | Identifies how to read 'Config' data from the @.toml@ file.
@@ -129,6 +131,7 @@ configT = Config
     <*> decision        "bench"       .= cBench
     <*> lastT (Toml.table preludeT) "prelude" .= cPrelude
     <*> extensions      "extensions"  .= cExtensions
+    <*> warnings        "warnings"    .= cWarnings
   where
     lastT :: (Key -> BiToml a) -> Key -> BiToml (Last a)
     lastT f = dimap getLast Last . Toml.maybeT f
@@ -168,6 +171,9 @@ configT = Config
         <$> Toml.text "package" .= cpPackage
         <*> Toml.text "module"  .= cpModule
 
+    warnings :: Key -> BiToml [Text]
+    warnings = dimap Just maybeToMonoid . Toml.maybeT (Toml.arrayOf Toml._Text)
+
 -- | Make sure that all the required configurations options were specified.
 finalise :: PartialConfig -> Validation [Text] Config
 finalise Config{..} = Config
@@ -189,6 +195,7 @@ finalise Config{..} = Config
     <*> pure cBench
     <*> pure cPrelude
     <*> pure cExtensions
+    <*> pure cWarnings
   where
     fin name = maybe (Failure ["Missing field: " <> name]) Success . getLast
 
--- a/src/Summoner/Template.hs
+++ b/src/Summoner/Template.hs
@@ -109,18 +109,37 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
         $endLine
         |]
 
+    createWarnings :: Text
+    createWarnings = case warnings of
+        [] -> defaultWarnings
+        xs -> "ghc-options:        " <> T.intercalate "\n                    " xs
+
+    defaultWarnings :: Text
+    defaultWarnings =
+        [text|
+        ghc-options:        -Wall
+                            -threaded
+                            -rtsopts
+                            -with-rtsopts=-N
+                            -Wincomplete-uni-patterns
+                            -Wincomplete-record-updates
+                            -Wmissing-import-lists
+                            -Wcompat
+                            -Widentities
+        |]
+
     createCabalLib :: Text
     createCabalLib =
         [text|
         library
-          hs-source-dirs:      src
-          exposed-modules:     $libModuleName
-                               $preludeMod
-          ghc-options:         -Wall
-          build-depends:       $base
-                             $customPreludePack
-          default-language:    Haskell2010
-          $defaultExtensions
+            hs-source-dirs:     src
+            exposed-modules:    $libModuleName
+                                $preludeMod
+            $createWarnings
+            build-depends:      $base
+                              $customPreludePack
+            default-language:   Haskell2010
+            $defaultExtensions
         $endLine
         |]
 
@@ -128,14 +147,14 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
     createCabalExe r =
         [text|
         executable $repo
-          hs-source-dirs:      app
-          main-is:             Main.hs
-          ghc-options:         -Wall -threaded -rtsopts -with-rtsopts=-N
-          build-depends:       $base
-                             $r
-                             $customPreludePack
-          default-language:    Haskell2010
-          $defaultExtensions
+            hs-source-dirs:     app
+            main-is:            Main.hs
+            $createWarnings
+            build-depends:      $base
+                              $r
+                              $customPreludePack
+            default-language:   Haskell2010
+            $defaultExtensions
         $endLine
         |]
 
@@ -143,15 +162,15 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
     createCabalTest =
         [text|
         test-suite ${repo}-test
-          type:                exitcode-stdio-1.0
-          hs-source-dirs:      test
-          main-is:             Spec.hs
-          build-depends:       $base
-                             , $repo
-                             $customPreludePack
-          ghc-options:         -Wall -Werror -threaded -rtsopts -with-rtsopts=-N
-          default-language:    Haskell2010
-          $defaultExtensions
+            type:                exitcode-stdio-1.0
+            hs-source-dirs:      test
+            main-is:             Spec.hs
+            build-depends:       $base
+                               , $repo
+                               $customPreludePack
+            $createWarnings
+            default-language:   Haskell2010
+            $defaultExtensions
         $endLine
         |]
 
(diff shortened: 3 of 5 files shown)


willbasky:
I didn`t add three warnings yet
```
-Wredundant-constraints (since 8.0)
-Wmissing-export-lists (since 8.4.1)
-Wpartial-fields (since 8.4.1)
```
Coming soon.

src/Summoner/Config.hs
```diff
@@ -168,6 +171,9 @@ configT = Config
         <$> Toml.text "package" .= cpPackage
         <*> Toml.text "module"  .= cpModule
 
+    warnings :: Key -> BiToml [Text]
+    warnings = dimap Just maybeToMonoid . Toml.maybeT (Toml.arrayOf Toml._Text)
```
vrom911: We already have this `BiToml` for the list of `Text`, it can be reused 🙂 You can even rename it accordingly so its name won't confused when used with warnings as well.

src/Summoner/Template.hs
```diff
@@ -109,66 +109,96 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
+
+    warningFromGHC8 :: Text
+    warningFromGHC8 =
+        if Ghc7103 `elem` testedVersions then ""
+        else
+            [text|
+            -Wredundant-constraints
+            -Wmissing-export-lists
+            -Wpartial-fields
+            |]
```
vrom911: Very nice solution! 👍

src/Summoner/Template.hs
```diff
@@ -109,66 +109,96 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
+        [] -> defaultWarnings
+        xs -> "ghc-options:        " <> T.intercalate "\n                    " xs
+
+    defaultWarnings :: Text
+    defaultWarnings =
+        [text|
+        ghc-options:        -Wall
+                            -threaded
+                            -rtsopts
+                            -with-rtsopts=-N
```
vrom911: `-threaded` `-rtsopts` and `-with-rtsopts` should be in all of the `ghc-options` fields except library stanza no matter what.

src/Summoner/Template.hs
```diff
@@ -109,66 +109,96 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
-          default-language:    Haskell2010
-          $defaultExtensions
+            hs-source-dirs:     src
+            exposed-modules:    $libModuleName
+                                $preludeMod
+            $createWarnings
+            build-depends:      $base
+                              $customPreludePack
+            default-language:   Haskell2010
+            $defaultExtensions
```
vrom911: Can you please bring back indentation? We use 2 in the cabal file. And we actually have all the values aligned and these changes will break it :(

src/Summoner/Template.hs
ChShersh: Also, let's add `-fhide-source-paths` if GHC >= 8.2.2

src/Summoner/Template.hs
```diff
@@ -109,66 +109,96 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
-          $defaultExtensions
+            type:               exitcode-stdio-1.0
+            default-language:   Haskell2010
+            $createWarnings
+            hs-source-dirs:     benchmark
+            main-is:            Main.hs
+            build-depends:      $base
+                              , gauge
+                              $customPreludePack
+                              $r
```
ChShersh: I just noticed that `benchmarks` and `executable` has `r` variable. So the code can be improved even further. Could you, please, do the following?

1. Swap lines with `$customPreludePack` and `$r`
2. Add `r` variable to `createCabalTest` function?

ChShersh:
> Benchmark field had -02 warning. Is it necessary?

@willbasky Yes, `benchmarks` should have `-02` enabled. Also, it's not a warning. It's an option that controls optimization level.

src/Summoner/Template.hs
```diff
@@ -109,6 +111,34 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
+        -Wcompat
+        -Widentities
+        $versionWarnings
+        |]
+
+    versionWarnings :: Text
+    versionWarnings
+        =  memptyIfFalse (Ghc7103 `notElem` testedVersions)
+            "-Wredundant-constraints\n"
+        <> memptyIfFalse (null (testedVersions \\ [Ghc822, Ghc843]))
```
ChShersh: This implementation doesn't scale. Imagine that GHC-8.6.1 released. And now we need not forget to update 
 all such places. It's recommended to write code in a such way that minimizes amount of things needed to be remembered. It's better if you implement function like:

```haskell
hasAtLeast :: Ord a => [a] -> a -> Bool
```

That checks that there's at least one element that greater than given element.

And then you can write only once:

```haskell
meptyIfFalse (testedVersions `hasAtLeast` Ghc822)
    "some text"
```

src/Summoner/Template.hs
```diff
@@ -117,8 +147,9 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
           exposed-modules:     $libModuleName
                                $preludeMod
           ghc-options:         -Wall
+                               $ghcOptions
           build-depends:       $base
-                             $customPreludePack
+                               $customPreludePack
```
ChShersh: You don't need to add extra 2 spaces here. `customPreludePack` variable contains `", "` at the beginning

src/Summoner/Template.hs
```diff
@@ -127,10 +158,14 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
     createCabalExe :: Text -> Text
     createCabalExe r =
         [text|
-        executable $repo
+        executable             $repo
```
ChShersh: Alignment after `executable` keyword is not needed

src/Summoner/Template.hs
```diff
@@ -139,17 +174,22 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
         $endLine
         |]
 
-    createCabalTest :: Text
-    createCabalTest =
+    createCabalTest :: Text -> Text
+    createCabalTest r =
         [text|
-        test-suite ${repo}-test
+        test-suite             ${repo}-test
```
ChShersh: Alignment after `test-suite` keyword is not needed

src/Summoner/Template.hs
```diff
@@ -127,10 +158,14 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
         [text|
-        executable $repo
+        executable             $repo
           hs-source-dirs:      app
           main-is:             Main.hs
-          ghc-options:         -Wall -threaded -rtsopts -with-rtsopts=-N
+          ghc-options:         -Wall
+                               -threaded
+                               -rtsopts
+                               -with-rtsopts=-N
```
ChShersh: Nice idea to put all options on separate line!

src/Summoner/Template.hs
```diff
@@ -109,6 +111,34 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
         $endLine
         |]
 
+    ghcOptions :: Text
+    ghcOptions = case warnings of
+        [] -> extraWarnings
```
ChShersh: Better name this variable as it was: `defaultWarnings`

src/Summoner/Template.hs
```diff
@@ -25,30 +25,32 @@ import qualified Data.Text as T
 -- Stack File Creation
 ----------------------------------------------------------------------------
 
-emptyIfNot :: Monoid m => Bool -> m -> m
-emptyIfNot p val = if p then val else mempty
+memptyIfFalse :: Monoid m => Bool -> m -> m
+memptyIfFalse p val = if p then val else mempty
```
ChShersh: Thanks a lot of this refactoring! :slightly_smiling_face:

src/Summoner/Template.hs
```diff
@@ -139,17 +174,22 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
+    createCabalTest r =
         [text|
-        test-suite ${repo}-test
+        test-suite             ${repo}-test
           type:                exitcode-stdio-1.0
           hs-source-dirs:      test
           main-is:             Spec.hs
           build-depends:       $base
                              , $repo
+                             $r
```
ChShersh: You don't need `, $repo` if you have `$r`, The idea is that if there's no `library` stanza in project, we don't want to add dependency for this library.

src/Summoner/Template.hs
```diff
@@ -158,16 +198,21 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
-          type:                exitcode-stdio-1.0
-          default-language:    Haskell2010
-          ghc-options:         -Wall -Werror -O2 -threaded -rtsopts -with-rtsopts=-N
-          hs-source-dirs:      benchmark
-          main-is:             Main.hs
-          build-depends:       $base
-                             , gauge
-                             $customPreludePack
-                             $r
+        benchmark             ${repo}-benchmark
```
ChShersh: Alignment here in this line after `benchmark` keyword is not needed

src/Summoner/Template.hs
```diff
@@ -158,16 +198,21 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
+          type:               exitcode-stdio-1.0
+          default-language:   Haskell2010
```
ChShersh: Let's move `default-language` down right before `default-extensions`

src/Summoner/Template.hs
```diff
@@ -158,16 +198,21 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
+          ghc-options:        -Wall
+                              -threaded
+                              -rtsopts
+                              -with-rtsopts=-N
+                              -02
+                              $ghcOptions
+          hs-source-dirs:     benchmark
+          main-is:            Main.hs
```
ChShersh: Let's unify structure of stanzas. And let's have `hs-source-dirs` with `main-is` before `ghc-options` field how it's already done in executable an test-suite.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
   Fix travis cache directories
 * [#110](https://github.com/kowainik/summoner/issues/110):
   Print executing command
+* [#91](https://github.com/kowainik/summoner/issues/91):
+  Add warning fields for ghc
```
ChShersh: @willbasky Could you also please update documentation in `README` and tell there that we have `warnings` field in configuration?

src/Summoner/Template.hs
```diff
@@ -109,6 +111,37 @@ createStackTemplate ProjectData{..} = Dir (toString repo) $
...
+        -Wincomplete-uni-patterns
+        -Wincomplete-record-updates
+        -Wmissing-import-lists
+        -Wcompat
+        -Widentities
+        $versionWarnings
+        |]
+
+    hasLeast :: Ord a => [a] -> a -> Bool
+    hasLeast list el = list == filter (>= el) list
```
ChShersh: The implementation of this function is wrong. If I call this function like this:

```haskell
hasAtLeast [Ghc7103, Ghc843] Ghc822
```

then it will return `False` while it should return `True`. `list` is `[Ghc7103, Ghc843]` and `filter (>= el) list` is `[Ghc843]`. Obviously, `[Ghc7103, Ghc843] /= [Ghc843]` and that's why we get `False`.

README.md
```diff
@@ -216,6 +216,25 @@ project-name
 ```
 and also repository with one commit at master will be added with enabled `Travis-CI` for that.
 
+## Warnings
+
+`ghc-options` field has default warnings in the  _project-name.cabal_:
```
ChShersh: Strictly speaking, not all of these are warnings. It's better to tell something like this:

The following warning checks are added by default to every stanza:

-Wall

The following `ghc-options` are added to `executable`, `test-suite` and `benchmark` stanza.

-threaded and other friends

Also, you need to update part of README regarding `tomland` configuration and tell users that there's `warnings` field in toml file. And that users can override set of warnings through configuration.

Issue #90, opened 2018-06-03
Write more property tests for TOML


ChShersh:
Ideas for more tests:

- [ ] `decode . encode configT ≡ Right` — property test for our toml config. Should be guaranteed by `tomland` unless we did some mistakes in our config.

PR #119, opened 2018-08-06
[#90] Add TOML test

willbasky:
Resolves [#90](https://github.com/kowainik/summoner/issues/90)

### DIFF
--- a/summoner.cabal
+++ b/summoner.cabal
@@ -91,6 +91,7 @@ test-suite summoner-test
                      , relude
                      , tasty
                      , tasty-hedgehog
+                     , tomland
                      , summoner
   default-extensions:  NoImplicitPrelude
 
--- a/summoner.cabal
+++ b/summoner.cabal
@@ -44,6 +44,12 @@ library
   autogen-modules:     Paths_summoner
 
   ghc-options:         -Wall
+                       -Wincomplete-uni-patterns
+                       -Wincomplete-record-updates
+                       -Wcompat
+                       -Widentities
+                       -Wredundant-constraints
+                       -fhide-source-paths
   build-depends:       base >= 4.9 && < 5
                      , aeson
                      , ansi-terminal
@@ -72,7 +78,16 @@ library
 executable summon
   hs-source-dirs:      app
   main-is:             Main.hs
-  ghc-options:         -Wall -threaded -rtsopts -with-rtsopts=-N
+  ghc-options:         -Wall 
+                       -threaded  
+                       -rtsopts 
+                       -with-rtsopts=-N
+                       -Wincomplete-uni-patterns
+                       -Wincomplete-record-updates
+                       -Wcompat
+                       -Widentities
+                       -Wredundant-constraints
+                       -fhide-source-paths
   build-depends:       base
                      , relude
                      , summoner
@@ -84,6 +99,7 @@ test-suite summoner-test
 
   main-is:             Spec.hs
   other-modules:       Test.DecisionSpec
+                       Test.TomlSpec
 
   build-tool-depends:  tasty-discover:tasty-discover
   build-depends:       base
@@ -95,5 +111,14 @@ test-suite summoner-test
                      , summoner
   default-extensions:  NoImplicitPrelude
 
-  ghc-options:         -Wall -threaded -rtsopts -with-rtsopts=-N
+  ghc-options:         -Wall 
+                       -threaded 
+                       -rtsopts 
+                       -with-rtsopts=-N
+                       -Wincomplete-uni-patterns
+                       -Wincomplete-record-updates
+                       -Wcompat
+                       -Widentities
+                       -Wredundant-constraints
+                       -fhide-source-paths
   default-language:    Haskell2010
--- a/test/Test/DecisionSpec.hs
+++ b/test/Test/DecisionSpec.hs
@@ -1,6 +1,8 @@
-module Test.DecisionSpec where
+module Test.DecisionSpec
+    (genDecision)
+    where
 
-import Relude (mempty, pure, ($), (<>))
+import Relude
 
 import Hedgehog (MonadGen, forAll, property, (===))
 import Test.Tasty (TestTree)
--- a/test/Test/DecisionSpec.hs
+++ b/test/Test/DecisionSpec.hs
@@ -5,6 +5,7 @@ import Relude
 import Hedgehog (MonadGen, forAll, liftGen, property, (===))
 import Test.Tasty (TestTree)
 import Test.Tasty.Hedgehog (testProperty)
+import Toml.Bi.Code (decode, encode, prettyException)
 
 import Summoner.Config (ConfigP (..), PartialConfig, configT)
 import Summoner.License (License (..), licenseNames)
@@ -12,7 +13,6 @@ import Summoner.ProjectData (CustomPrelude (..), Decision, GhcVer (..))
 
 import qualified Hedgehog.Gen as Gen
 import qualified Hedgehog.Range as Range
--- import qualified Toml
 
 test_decisionSemigroupAssoc :: [TestTree]
 test_decisionSemigroupAssoc = pure $ testProperty "Decision Semigroup:Assoc" $ property $ do
@@ -51,9 +51,9 @@ randomConfig :: MonadGen m => m PartialConfig
 randomConfig = do
     text <- liftGen genText
     list <- liftGen genTextArr
-    dec <- liftGen genDecision
-    lic <- liftGen genLicense
-    ghc <- liftGen genGhcVerArr
+    dec  <- liftGen genDecision
+    lic  <- liftGen genLicense
+    ghc  <- liftGen genGhcVerArr
     lude <- liftGen genCustomPrelude
     pure Config
         { cOwner      = Last $ Just text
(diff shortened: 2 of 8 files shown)


ChShersh:
@willbasky If PR is unfinished, you can add label `wip` to it. And write `[WIP]` at the beginning of the PR title. So everyone will see that PR is WIP and won't try to review it until it's not WIP.

summoner.cabal
```diff
@@ -91,6 +91,7 @@ test-suite summoner-test
                      , relude
                      , tasty
                      , tasty-hedgehog
+                     , tomland
                      , summoner
   default-extensions:  NoImplicitPrelude
```
ChShersh: @willbasky One more thing: in your previous PR you've introduced default warnings to generated `.cabal` files. Could you also add those warnings to `summoner.cabal` file and fix any appear warnings? I think we should test warnings that we generated. It probably should be done under previous PR but let's do this here.

test/Test/DecisionSpec.hs
```diff
@@ -25,3 +29,59 @@ test_decisionMonoidMempty = pure $ testProperty "Decision Monoid:Mempty" $ prope
...
+        , cExe        = dec
+        , cTest       = dec
+        , cBench      = dec
+        , cPrelude    = Last $ Just lude
+        , cExtensions = list
+        , cWarnings   = list
+        }
+
+test_Toml :: [TestTree]
+test_Toml = pure $ testProperty "Test TOML" $ property $ do
```
ChShersh: This test should be in it's own module. I guess `Test.TomlSpec` is an okay name.

test/Test/DecisionSpec.hs
```diff
@@ -25,3 +29,59 @@ test_decisionMonoidMempty = pure $ testProperty "Decision Monoid:Mempty" $ prope
...
+    pure Config
+        { cOwner      = Last $ Just text
+        , cFullName   = Last $ Just text
+        , cEmail      = Last $ Just text
+        , cLicense    = Last $ Just lic
+        , cGhcVer     = Last $ Just ghc
+        , cCabal      = dec
+        , cStack      = dec
+        , cGitHub     = dec
+        , cTravis     = dec
```
ChShersh: This is not a real test, because all fields will have the same value. I think it's better if you use `RecordWildCards` and write code like this:

```haskell
do
    cOwner <- Last . Just <$> genText
    cCabal <- genDecision
    ...
    pure Config{..}
```

test/Test/DecisionSpec.hs
```diff
@@ -25,3 +29,59 @@ test_decisionMonoidMempty = pure $ testProperty "Decision Monoid:Mempty" $ prope
...
+        }
+
+test_Toml :: [TestTree]
+test_Toml = pure $ testProperty "Test TOML" $ property $ do
+    configToml     <- forAll randomConfig
+    let textToml    = encode configT configToml
+    let mConfigToml = decode configT textToml
+    case mConfigToml of
+        Left err           -> putStrLn $ prettyException err
+        Right deConfigToml -> configToml === deConfigToml
```
ChShersh: It's better to use `tripping` function from `hedgehog`. It provides better error message:

* https://hackage.haskell.org/package/hedgehog-0.6/docs/Hedgehog.html#v:tripping

test/Test/DecisionSpec.hs
ChShersh: Let's write some message here like:

```haskell
testPropery "decode . encode == id
```

test/Test/DecisionSpec.hs
```diff
@@ -25,3 +29,59 @@ test_decisionMonoidMempty = pure $ testProperty "Decision Monoid:Mempty" $ prope
...
+genCustomPrelude :: MonadGen m => m CustomPrelude
+genCustomPrelude = Prelude
+    <$> Gen.text (Range.constant 0 10) Gen.alpha
+    <*> Gen.text (Range.constant 0 10) Gen.alpha
+
+genLicense :: MonadGen m => m License
+genLicense = Gen.element licenseNames
+
+randomConfig :: MonadGen m => m PartialConfig
+randomConfig = do
```
ChShersh: Let's use naming convention and name this function as `genPartialConfig`.

test/Test/DecisionSpec.hs
```diff
@@ -25,3 +29,59 @@ test_decisionMonoidMempty = pure $ testProperty "Decision Monoid:Mempty" $ prope
...
+genText :: MonadGen m => m Text
+genText = Gen.text (Range.constant 0 10) Gen.alpha
+
+genGhcVerArr :: MonadGen m => m [GhcVer]
+genGhcVerArr = Gen.list (Range.constant 0 10) Gen.enumBounded
+
+genCustomPrelude :: MonadGen m => m CustomPrelude
+genCustomPrelude = Prelude
+    <$> Gen.text (Range.constant 0 10) Gen.alpha
+    <*> Gen.text (Range.constant 0 10) Gen.alpha
```
ChShersh: I think it's not only `alpha`. Package names can contain characters (uppercase and lowercase), digits, `-`.
Module names can contain characters (uppercase and lowercase), digits and dots.

test/Test/DecisionSpec.hs
```diff
@@ -25,3 +29,59 @@ test_decisionMonoidMempty = pure $ testProperty "Decision Monoid:Mempty" $ prope
 
 genDecision :: MonadGen m => m Decision
 genDecision = Gen.enumBounded
+
+genTextArr :: MonadGen m => m [Text]
+genTextArr = Gen.list (Range.constant 0 10) $ Gen.text (Range.constant 0 10) Gen.alpha
```
ChShersh: I think you can reuse `genText` function here.

test/Test/DecisionSpec.hs
willbasky: Will values be generated newly every time in that `<-`  case?

test/Test/DecisionSpec.hs
```diff
@@ -1,6 +1,6 @@
 module Test.DecisionSpec where
 
-import Relude
```
ChShersh: I think you can return this back to implicit import

test/Test/TomlSpec.hs
```diff
@@ -0,0 +1,67 @@
...
+genTextArr :: MonadGen m => m [Text]
+genTextArr = Gen.list (Range.constant 0 10) genText
+
+genGhcVerArr :: MonadGen m => m [GhcVer]
+genGhcVerArr = Gen.list (Range.constant 0 10) Gen.enumBounded
+
+genCustomPrelude :: MonadGen m => m CustomPrelude
+genCustomPrelude = Prelude
+    <$> genText
+    <*> genText
```
ChShersh: I think it's okay to write this function in single line. Usually splitting to multiple lines make sense when the line becomes too big.

test/Test/TomlSpec.hs
```diff
@@ -0,0 +1,67 @@
...
+    cTest       <- genDecision
+    cBench      <- genDecision
+    cPrelude    <- Last . Just <$> genCustomPrelude
+    cExtensions <- genTextArr
+    cWarnings   <- genTextArr
+    pure Config{..}
+
+test_Toml :: [TestTree]
+test_Toml = pure $ testProperty "decode . encode == id" $ property $ do
+    configToml     <- forAll genPartialConfig
```
ChShersh: I think alignment here is redundant.

test/Test/TomlSpec.hs
```diff
@@ -0,0 +1,67 @@
...
+
+import Summoner.Config (ConfigP (..), PartialConfig, configT)
+import Summoner.License (License (..), licenseNames)
+import Summoner.ProjectData (CustomPrelude (..), Decision, GhcVer (..))
+
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+genDecision :: MonadGen m => m Decision
+genDecision = Gen.enumBounded
```
ChShersh: I think you don't need to write `genDecision` here. You can just import it from `Test.DecisionSpec`.

test/Test/TomlSpec.hs
```diff
@@ -0,0 +1,67 @@
+{-# LANGUAGE RecordWildCards #-}
+
+module Test.TomlSpec where
+
+import Relude (Last (Last), Maybe (Just), Text, pure, ($), (.), (<$>), (<*>))
```
ChShersh: Looks like this import also doesn't need to be explicit.

summoner.cabal
```diff
@@ -44,6 +44,12 @@ library
   autogen-modules:     Paths_summoner
 
   ghc-options:         -Wall
+                       -Wincomplete-uni-patterns
+                       -Wincomplete-record-updates
+                       -Wcompat
+                       -Widentities
+                       -Wredundant-constraints
+                       -fhide-source-paths
```
ChShersh: Wow, I'm surprised that we added these warnings to `library` and we don't need to fix anything! Our code is already perfect ✨

ChShersh:
This looks really great so far!

test/Test/DecisionSpec.hs
```diff
@@ -1,6 +1,8 @@
-module Test.DecisionSpec where
+module Test.DecisionSpec
+    (genDecision)
+    where
```
ChShersh: This export is not formatted properly. And moreover, now tests from this module are not running since you don't export them.

Issue #58, opened 2018-05-23
Make `License` type safer


vrom911:
Take inspiration from `GhcVer` data type.

PR #123, opened 2018-08-10
[#58] Make License type safer

willbasky:
Resolves [#58](https://github.com/kowainik/summoner/issues/58)

### DIFF
--- a/src/Summoner/Config.hs
+++ b/src/Summoner/Config.hs
@@ -33,7 +33,7 @@ import Generics.Deriving.Monoid (GMonoid, gmemptydefault)
 import Generics.Deriving.Semigroup (GSemigroup, gsappenddefault)
 import Toml (AnyValue (..), BiToml, Key, Prism (..), dimap, (.=))
 
-import Summoner.License (LicenseName (..), parseLicense, showLicense)
+import Summoner.License (LicenseName (..), cabalLicenseName, parseLicense)
 import Summoner.ProjectData (CustomPrelude (..), Decision (..), GhcVer (..), parseGhcVer,
                              showGhcVer)
 import Summoner.Validation (Validation (..))
@@ -66,23 +66,22 @@ data ConfigP (p :: Phase) = Config
     , cWarnings   :: [Text]
     } deriving (Generic)
 
-deriving instance 
+deriving instance
     ( GSemigroup (p :- Text)
     , GSemigroup (p :- LicenseName)
     , GSemigroup (p :- [GhcVer])
     ) => GSemigroup (ConfigP p)
-deriving instance 
+deriving instance
     ( GMonoid (p :- Text)
     , GMonoid (p :- LicenseName)
     , GMonoid (p :- [GhcVer])
     ) => GMonoid (ConfigP p)
-deriving instance 
+deriving instance
     ( Eq (p :- Text)
     , Eq (p :- LicenseName)
     , Eq (p :- [GhcVer])
-    , Eq (Last CustomPrelude)
     ) => Eq (ConfigP p)
-deriving instance 
+deriving instance
     ( Show (p :- Text)
     , Show (p :- LicenseName)
     , Show (p :- [GhcVer])
@@ -154,7 +153,7 @@ configT = Config
     <*> textArr         "warnings"    .= cWarnings
   where
     lastT :: (Key -> BiToml a) -> Key -> BiToml (Last a)
-    lastT f = dimap getLast Last . Toml.maybeT f
+    lastT = Toml.wrapper . Toml.maybeT
 
     _GhcVer :: Prism AnyValue GhcVer
     _GhcVer = Prism
@@ -166,7 +165,7 @@ configT = Config
     ghcVerArr = Toml.arrayOf _GhcVer
 
     license :: Key -> BiToml LicenseName
-    license = Toml.mdimap showLicense parseLicense . Toml.text
+    license = Toml.mdimap cabalLicenseName parseLicense . Toml.text
 
     textArr :: Key -> BiToml [Text]
     textArr = dimap Just maybeToMonoid . Toml.maybeT (Toml.arrayOf Toml._Text)
--- a/src/Summoner/Config.hs
+++ b/src/Summoner/Config.hs
@@ -33,7 +33,7 @@ import Generics.Deriving.Monoid (GMonoid, gmemptydefault)
 import Generics.Deriving.Semigroup (GSemigroup, gsappenddefault)
 import Toml (AnyValue (..), BiToml, Key, Prism (..), dimap, (.=))
 
-import Summoner.License (LicenseName (..), cabalLicenseName, parseLicense)
+import Summoner.License (LicenseName (..), parseLicense)
 import Summoner.ProjectData (CustomPrelude (..), Decision (..), GhcVer (..), parseGhcVer,
                              showGhcVer)
 import Summoner.Validation (Validation (..))
@@ -165,7 +165,7 @@ configT = Config
     ghcVerArr = Toml.arrayOf _GhcVer
 
     license :: Key -> BiToml LicenseName
-    license = Toml.mdimap cabalLicenseName parseLicense . Toml.text
+    license = Toml.mdimap show parseLicense . Toml.text
 
     textArr :: Key -> BiToml [Text]
     textArr = dimap Just maybeToMonoid . Toml.maybeT (Toml.arrayOf Toml._Text)
--- a/src/Summoner/Config.hs
+++ b/src/Summoner/Config.hs
@@ -33,7 +33,7 @@ import Generics.Deriving.Monoid (GMonoid, gmemptydefault)
 import Generics.Deriving.Semigroup (GSemigroup, gsappenddefault)
 import Toml (AnyValue (..), BiToml, Key, Prism (..), dimap, (.=))
 
-import Summoner.License (License (..), parseLicense, showLicense)
+import Summoner.License (LicenseName (..), parseLicense, showLicense)
 import Summoner.ProjectData (CustomPrelude (..), Decision (..), GhcVer (..), parseGhcVer,
                              showGhcVer)
 import Summoner.Validation (Validation (..))
@@ -48,7 +48,7 @@ data ConfigP (p :: Phase) = Config
     { cOwner      :: p :- Text
     , cFullName   :: p :- Text
     , cEmail      :: p :- Text
-    , cLicense    :: p :- License
+    , cLicense    :: p :- LicenseName
     , cGhcVer     :: p :- [GhcVer]
     , cCabal      :: Decision
     , cStack      :: Decision
@@ -66,10 +66,27 @@ data ConfigP (p :: Phase) = Config
     , cWarnings   :: [Text]
     } deriving (Generic)
 
-deriving instance (GSemigroup (p :- Text), GSemigroup (p :- License), GSemigroup (p :- [GhcVer])) => GSemigroup (ConfigP p)
-deriving instance (GMonoid (p :- Text), GMonoid (p :- License), GMonoid (p :- [GhcVer])) => GMonoid (ConfigP p)
-deriving instance (Eq (p :- Text), Eq (p :- License), Eq (p :- [GhcVer]), Eq (Last CustomPrelude)) => Eq (ConfigP p)
-deriving instance (Show (p :- Text), Show (p :- License), Show (p :- [GhcVer])) => Show (ConfigP p)
+deriving instance 
+    ( GSemigroup (p :- Text)
+    , GSemigroup (p :- LicenseName)
+    , GSemigroup (p :- [GhcVer])
+    ) => GSemigroup (ConfigP p)
+deriving instance 
+    ( GMonoid (p :- Text)
+    , GMonoid (p :- LicenseName)
+    , GMonoid (p :- [GhcVer])
+    ) => GMonoid (ConfigP p)
+deriving instance 
+    ( Eq (p :- Text)
+    , Eq (p :- LicenseName)
+    , Eq (p :- [GhcVer])
+    , Eq (Last CustomPrelude)
+    ) => Eq (ConfigP p)
+deriving instance 
+    ( Show (p :- Text)
+    , Show (p :- LicenseName)
+    , Show (p :- [GhcVer])
+    ) => Show (ConfigP p)
 
 infixl 3 :-
 type family phase :- field where
@@ -148,7 +165,7 @@ configT = Config
     ghcVerArr :: Key -> BiToml [GhcVer]
     ghcVerArr = Toml.arrayOf _GhcVer
 
-    license :: Key -> BiToml License
+    license :: Key -> BiToml LicenseName
     license = Toml.mdimap showLicense parseLicense . Toml.text
 
     textArr :: Key -> BiToml [Text]
--- a/src/Summoner/Config.hs
+++ b/src/Summoner/Config.hs
@@ -33,7 +33,7 @@ import Generics.Deriving.Monoid (GMonoid, gmemptydefault)
 import Generics.Deriving.Semigroup (GSemigroup, gsappenddefault)
 import Toml (AnyValue (..), BiToml, Key, Prism (..), dimap, (.=))
 
-import Summoner.License (License (..))
+import Summoner.License (License (..), parseLicense, showLicense)
 import Summoner.ProjectData (CustomPrelude (..), Decision (..), GhcVer (..), parseGhcVer,
                              showGhcVer)
 import Summoner.Validation (Validation (..))
@@ -95,7 +95,7 @@ defaultConfig = Config
     { cOwner    = Last (Just "kowainik")
     , cFullName = Last (Just "Kowainik")
     , cEmail    = Last (Just "xrom.xkov@gmail.com")
-    , cLicense  = Last (Just $ License "MIT")
+    , cLicense  = Last (Just MIT)
     , cGhcVer   = Last (Just [])
     , cCabal    = Idk
     , cStack    = Idk
@@ -149,7 +149,7 @@ configT = Config
     ghcVerArr = Toml.arrayOf _GhcVer
 
     license :: Key -> BiToml License
-    license =  dimap unLicense License . Toml.text
+    license = Toml.mdimap showLicense parseLicense . Toml.text
 
     textArr :: Key -> BiToml [Text]
     textArr = dimap Just maybeToMonoid . Toml.maybeT (Toml.arrayOf Toml._Text)
--- a/src/Summoner/Config.hs
+++ b/src/Summoner/Config.hs
@@ -33,7 +33,7 @@ import Generics.Deriving.Monoid (GMonoid, gmemptydefault)
 import Generics.Deriving.Semigroup (GSemigroup, gsappenddefault)
 import Toml (AnyValue (..), BiToml, Key, Prism (..), dimap, (.=))
 
-import Summoner.License (LicenseName (..), parseLicense)
+import Summoner.License (LicenseName (..))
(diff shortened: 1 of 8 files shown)


test/Test/TomlSpec.hs
```diff
@@ -3,14 +3,15 @@
...
+import Relude.Extra.Enum (universe)
 
 import Hedgehog (MonadGen, forAll, property, tripping)
 import Test.Tasty (TestTree)
 import Test.Tasty.Hedgehog (testProperty)
 import Toml.Bi.Code (decode, encode)
 
 import Summoner.Config (ConfigP (..), PartialConfig, configT)
-import Summoner.License (License (..), licenseNames)
+import Summoner.License (LicenseName (..))
```
ghallak: This can be changed to `LiceseName` instead of `LicenseName (..)`.

src/Summoner/License.hs
```diff
@@ -15,25 +18,47 @@ import qualified Data.Text as T
...
+showLicense LGPL21   = "LGPL-2.1"   
+showLicense LGPL3    = "LGPL-3"     
+showLicense AGPL3    = "AGPL-3"     
+showLicense Apache20 = "Apache-2.0" 
+showLicense MPL20    = "MPL-2.0"
+
+parseLicense :: Text -> Maybe LicenseName
+parseLicense = inverseMap showLicense
+
+newtype License = License { unBody :: Text }
```
ghallak: It's better to keep this as `unLicense` instead of `unBody`.

src/Summoner/Project.hs
```diff
@@ -19,7 +19,7 @@ import System.Process (readProcess)
 import Summoner.Ansi (errorMessage, infoMessage, successMessage)
 import Summoner.Config (Config, ConfigP (..))
 import Summoner.Default (currentYear, defaultGHC)
-import Summoner.License (License (..), customizeLicense, githubLicenseQueryNames, licenseNames)
+import Summoner.License (License (..), customizeLicense, showLicense, githubLicenseQueryNames, parseLicense)
```
ghallak: These imports are not sorted. Could you please run `stylish-haskell` on the files that you've changed, and it will sort them for you.

src/Summoner/License.hs
vrom911: This can be `unLicense` 🙂

src/Summoner/License.hs
```diff
@@ -15,25 +18,47 @@ import qualified Data.Text as T
...
-    , ("LGPL-2.1",   "lgpl-2.1")
-    , ("LGPL-3",     "lgpl-3.0")
-    , ("AGPL-3",     "agpl-3.0")
-    , ("Apache-2.0", "apache-2.0")
-    , ("MPL-2.0",    "mpl-2.0")
-    ]
-
-newtype License = License { unLicense :: Text }
-    deriving (IsString, Eq, Ord, Show)
+data LicenseName = MIT
```
vrom911: Sorry, I'm not sure if we have this in our style guide, but I recently using this kind of formatting everywhere, How do you like that?

```haskell
data LicenceName
    = MIT
    | BSD2
    | ...
    deriving (...)
```

src/Summoner/Project.hs
```diff
@@ -51,12 +49,12 @@ generateProject projectName Config{..} = do
     email       <- queryDef "Maintainer e-mail: " cEmail
     putText categoryText
     category <- query "Category: "
-    license  <- choose "License: " $ map unLicense $ ordNub (cLicense : licenseNames)
+    license  <- choose "License: " $ map showLicense universe
```
vrom911: I guess the logic here changed a bit. 
We want in this list to have the default license (or the chosen one) to come first as it will be shown in brackets like `[MIT]/BSD-3/...`.

src/Summoner/Project.hs
```diff
@@ -51,12 +49,12 @@ generateProject projectName Config{..} = do
...
 
     -- License creation
-    let licenseGithub = snd
-                      $ Unsafe.head
-                      $ dropWhile ((/= license) . unLicense . fst) githubLicenseQueryNames
+    let licenseGithub = case parseLicense license of 
+            Just l -> githubLicenseQueryNames l
+            Nothing -> error "Unrecognised license name"
```
vrom911: This case shouldn't ever happen, but yeah, we can't prove that to compiler..

I'm now thinking about patching the `choose` function so it will take the list of `[a]` instead of `[Text]` and return the `a` (instead of `Text`). This will make the code safer and remove this unnecessary check. What do you think?

src/Summoner/Project.hs
willbasky: I moved to HIE and tried `brittany` formatter in vscode, but I couldn`t force it to work. So, `stylish-haskell` is back now.

src/Summoner/License.hs
willbasky: Yes, It is better. How I missed it!

src/Summoner/Config.hs
```diff
@@ -66,10 +66,27 @@ data ConfigP (p :: Phase) = Config
...
+deriving instance 
+    ( GMonoid (p :- Text)
+    , GMonoid (p :- LicenseName)
+    , GMonoid (p :- [GhcVer])
+    ) => GMonoid (ConfigP p)
+deriving instance 
+    ( Eq (p :- Text)
+    , Eq (p :- LicenseName)
+    , Eq (p :- [GhcVer])
+    , Eq (Last CustomPrelude)
```
ChShersh: That's strange that `Eq` requires 4 constraints, when all other instances requires only 3 :thinking:

src/Summoner/Project.hs
willbasky: It show like `[MIT]/BSD-3/...` still when summoner asks about license.

src/Summoner/License.hs
ghallak: @vrom911 Can this rule be added to [`.stylish-haskell`](https://github.com/kowainik/summoner/blob/master/.stylish-haskell.yaml)? I'm not sure if it's supported. :thinking:

src/Summoner/Project.hs
willbasky: > This case shouldn't ever happen, but yeah, we can't prove that to compiler..

It happened when I was wrong :)

src/Summoner/Config.hs
```diff
@@ -148,8 +165,8 @@ configT = Config
     ghcVerArr :: Key -> BiToml [GhcVer]
     ghcVerArr = Toml.arrayOf _GhcVer
```
ChShersh: Could you also change `lastT` parser couple lines about to use `wrapper` combinator from `tomland`?

* http://hackage.haskell.org/package/tomland-0.3.1/docs/Toml-Bi-Combinators.html#v:wrapper

src/Summoner/License.hs
vrom911: That would be very nice, but I don't see in the docs that it's capable to do that 😞

src/Summoner/License.hs
```diff
@@ -15,25 +18,47 @@ import qualified Data.Text as T
...
+             | LGPL21
+             | LGPL3
+             | AGPL3
+             | Apache20
+             | MPL20
+             deriving (Eq, Ord, Show, Enum, Bounded, Generic)
+
+githubLicenseQueryNames :: LicenseName -> Text
+githubLicenseQueryNames MIT      = "mit"
+githubLicenseQueryNames BSD2     = "bsd-2-clause"
```
ChShersh: This is a nice use-case for `-XLambdaCase` extension! When your function have very long name, you can write like this:

```haskell
githubLicenseQueryNames = \case
    MIT  -> "mit"
    BSD2 -> "bsd-2-clause"
    ...
```

src/Summoner/Project.hs
ghallak: I use [this plugin](https://marketplace.visualstudio.com/items?itemName=vigoo.stylish-haskell) for vscode and it's working for me, you could give it a try.

src/Summoner/License.hs
```diff
@@ -15,25 +18,47 @@ import qualified Data.Text as T
...
+showLicense MIT      = "MIT"
+showLicense BSD2     = "BSD2"       
+showLicense BSD3     = "BSD3"       
+showLicense GPL2     = "GPL-2"      
+showLicense GPL3     = "GPL-3"      
+showLicense LGPL21   = "LGPL-2.1"   
+showLicense LGPL3    = "LGPL-3"     
+showLicense AGPL3    = "AGPL-3"     
+showLicense Apache20 = "Apache-2.0" 
+showLicense MPL20    = "MPL-2.0"
```
ChShersh: It's better to name this function `cabalLicenseName`. It's main purpose is to use in `.cabal` file.

src/Summoner/License.hs
```diff
@@ -15,25 +18,47 @@ import qualified Data.Text as T
...
+
+parseLicense :: Text -> Maybe LicenseName
+parseLicense = inverseMap showLicense
```
ChShersh: :sweat_drops:

src/Summoner/License.hs
```diff
@@ -15,25 +18,47 @@ import qualified Data.Text as T
...
+
+newtype License = License { unBody :: Text }
+    deriving (IsString, Show, Generic)
 
 instance FromJSON License where
     parseJSON = withObject "License" $ \o -> License <$> o .: "body"
```
ChShersh: Actually, `customizeLicense` function can be refactored as well. Instead of

```haskell
l `elem` words "MIT BSD2 BSD3"
```

You can write:

```haskell
l `elem` [MIT, BSD2, BSD3]
```

src/Summoner/Project.hs
ChShersh: @willbasky Yeah, but when you try it with `.summoner.toml` from our org, it should show `MPL-2.0`. But it shows `MIT`. It shows MIT now, because it's the first constructor. And `universe` starts with `minBound`

* https://github.com/kowainik/org/blob/master/.summoner.toml

src/Summoner/Project.hs
ChShersh: I agree with @vrom911. I think it will be good if `choose` could take list of `a` and return `a` on success.

willbasky:
> I agree with @vrom911. I think it will be good if choose could take list of a and return a on success.

@ChShersh I doubted it well enough:
```
choose :: forall a. (Show a, IsString a) => Text -> [a] -> IO a
choose question choices = do
    let showChoices = map show choices
    printQuestion question showChoices
    answer <- prompt
    if | T.null answer -> pure (Unsafe.head choices)
       | answer `elem` showChoices -> pure $ fromString $ T.unpack answer
       | otherwise -> do
           errorMessage "This wasn't a valid choice."
           choose question choices
```

src/Summoner/License.hs
```diff
@@ -1,51 +1,81 @@
...
+    GPL2     -> "gpl-2.0"
+    GPL3     -> "gpl-3.0"
+    LGPL21   -> "lgpl-2.1"
+    LGPL3    -> "lgpl-3.0"
+    AGPL3    -> "agpl-3.0"
+    Apache20 -> "apache-2.0"
+    MPL20    -> "mpl-2.0"
+
+cabalLicenseName :: LicenseName -> Text
+cabalLicenseName = \case
```
vrom911: It's better to make this `show` function instead of useless derived instance :slightly_smiling_face:

src/Summoner/License.hs
```diff
@@ -1,51 +1,81 @@
...
+parseLicense = inverseMap cabalLicenseName
 
 newtype License = License { unLicense :: Text }
-    deriving (IsString, Eq, Ord, Show)
+    deriving (IsString, Show, Generic)
 
 instance FromJSON License where
     parseJSON = withObject "License" $ \o -> License <$> o .: "body"
 
 customizeLicense :: Text -> Text -> Text -> Text -> Text
```
vrom911: I guess it's safer to receive `LicenseName` in this function

src/Summoner/License.hs
```diff
@@ -1,51 +1,81 @@
...
 customizeLicense l t nm year
-    | l `elem` words "MIT BSD2 BSD3" = updateLicenseText
-    | otherwise = t
+    | l `elem` ["MIT", "BSD2", "BSD3"] = updateLicenseText
+    | otherwise                        = t
   where
     updateLicenseText =
```
vrom911: Would be nice to add top-level definition in here

src/Summoner/Question.hs
```diff
@@ -45,12 +45,13 @@ printQuestion question (def:rest) = do
     putTextLn $ "/" <> restSlash
 printQuestion question [] = T.putStrLn question
 
-choose :: Text -> [Text] -> IO Text
+choose :: (Show a, IsString a) => Text -> [a] -> IO a
```
vrom911: I guess here `IsString` won't work for enums. 
The solution that I can imagine is to make `Read a` constraint instead and then you can do `readmaybe @a answer` to decide what to do next. How do you think?

src/Summoner/License.hs
```diff
@@ -1,51 +1,81 @@
...
-       , githubLicenseQueryNames
-       ) where
+    ( LicenseName(..)
+    , License(..)
+    , cabalLicenseName
+    , customizeLicense
+    , githubLicenseQueryNames
+    , parseLicense
+    )
+where
```
ChShersh: Export list should be formatted according to style-guide:

* https://github.com/serokell/serokell-util/blob/master/serokell-style.md#export-lists

src/Summoner/License.hs
```diff
@@ -1,51 +1,81 @@
+{-# LANGUAGE LambdaCase #-}
```
ChShersh: You don't need to write extension here. It's already in `default-extensions` in `.cabal` file.

src/Summoner/License.hs
ChShersh: Yeah, I agree. Probably it's even better to have useful `Show` instance and don't have `cabalLicenseName` as a separate function.

src/Summoner/Question.hs
willbasky: I don`t mind. :)
> I guess here IsString won't work for enums.

Is it by reason of enum instance lack?

src/Summoner/Question.hs
```diff
@@ -45,13 +46,17 @@ printQuestion question (def:rest) = do
...
     let showChoices = map show choices
     printQuestion question showChoices
     answer <- prompt
     if | T.null answer -> pure (Unsafe.head choices)
-       | answer `elem` showChoices -> pure $ fromString $ T.unpack answer
+       | answer `elem` showChoices -> do
+           let mAnswer = readMaybe @a $ T.unpack answer
+           case mAnswer of
+               Nothing -> error "Unreadable answer type"
+               Just a  -> pure a
```
willbasky: Is it possible to make it better? Without `Nothing`?

vrom911: Well, I meant a little bit different think.
I want to get rid of `showChoices`. It's not needed at all to translate all to text and then back. What I propose if to do something like
```haskell
if T.null answer 
    then pure (Unsafe.head choices)
    else case readMaybe @a answer of
            Just a  -> pure a	       
            Nothing -> do
               errorMessage "This wasn't a valid choice."
               choose question choices
```

src/Summoner/Question.hs
```diff
@@ -45,15 +46,18 @@ printQuestion question (def:rest) = do
...
-    if | T.null answer -> pure (Unsafe.head choices)
-       | answer `elem` choices -> pure answer
-       | otherwise -> do
-           errorMessage "This wasn't a valid choice."
-           choose question choices
+    if T.null answer
+        then pure (Unsafe.head choices)
+        else case readMaybe @a (T.unpack answer) of
+                Just a  -> pure a
+                Nothing -> do
```
vrom911: Opps, sorry. Now I see that this is exactly `whenNothing` function

willbasky: We have found it!

src/Summoner/Config.hs
```diff
@@ -165,7 +165,7 @@ configT = Config
...
 
     license :: Key -> BiToml LicenseName
-    license = Toml.mdimap show parseLicense . Toml.text
```
ChShersh: It's okay to keep `parseLicense` here.

src/Summoner/License.hs
```diff
@@ -42,6 +43,23 @@ instance Show LicenseName where
     show Apache20 = "Apache-2.0"
     show MPL20    = "MPL-2.0"
 
+instance Read LicenseName where
+    readPrec = TR.choice $ tryParse
+        [ ("MIT", MIT), ("BSD2", BSD2)
```
ChShersh: You should implement `Read` instance using `parseLicense` function. The key point of `parseLicense` is to write mapping between `LicenseName` and `String` only once and then reuse it. This makes code more robust to refactoring.

src/Summoner/License.hs
```diff
@@ -44,15 +45,13 @@ instance Show LicenseName where
...
-        [ ("MIT", MIT), ("BSD2", BSD2)
-        , ("BSD3", BSD3), ("GPL-2", GPL2)
-        , ("GPL-3", GPL3), ("LGPL-2.1", LGPL21)
-        , ("LGPL-3", LGPL3), ("AGPL-3", AGPL3)
-        , ("Apache-2.0", Apache20), ("MPL-2.0", MPL20)
-        ]
-          where
-            tryParse = map (\(x, y) -> TR.lift $ string x >> return y)
+    readPrec = TR.parens
+        (do TR.Ident s <- TR.lexP
```
ChShersh: It's better if you write:

```haskel
readPrec = TR.parens $ do
   ...
```

No need to have `()`.

src/Summoner/License.hs
```diff
@@ -44,15 +45,13 @@ instance Show LicenseName where
...
+            let mLicenseName = parseLicenseName (T.pack s)
```
ChShersh: `T.pack` is `toText` from `relude`.

src/Summoner/License.hs
```diff
@@ -44,15 +45,13 @@ instance Show LicenseName where
...
+            case mLicenseName of
+                Just l  -> return l
+                Nothing -> TR.pfail
```
ChShersh: I think this case is `whenNothing mLicenseName TR.pfail`

ChShersh:
@willbasky Unfortunately, `instance Read` doesn't work properly. I'm observing the following behavior:

```haskell
λ: readMaybe @LicenseName "MIT"
Just MIT
λ: readMaybe @LicenseName "MPL-2.0"
Nothing
λ: parseLicenseName "MPL-2.0"
Just MPL-2.0
```

I guess we need to implement more complex logic in `Read` instance to take this into account.

Also, could you write roundtrip property-based tests for `Show/Read` for `LicenseName`? Since we have custom `Show` and `Read` instances, we want to be sure that they work properly for any values.

src/Summoner/Question.hs
```diff
@@ -46,14 +47,14 @@ printQuestion question (def:rest) = do
     putTextLn $ "/" <> restSlash
 printQuestion question [] = T.putStrLn question
 
-choose :: forall a. (Show a, Read a) => Text -> [a] -> IO a
+choose :: Text -> [LicenseName] -> IO LicenseName
```
vrom911: I would like to keep this function polymorphic, that will take a function how to parse the concrete type.

If you want to have a license specific one because you don't want to pass the function explicitly then you should create `chooseLicense = choose parseLicenseName`, but I don't see the problem in writing it down and bring `License` in this module.

ChShersh: Agree with @vrom911 here. It's better to have function that takes `(Text -> Maybe a)` function explicitly and not hardcode `LicenseName`.

ChShersh:
Works for me as well 👍

Issue #125, opened 2018-08-12
Split `Summoner.Template` into smaller pieces


ChShersh:
This is a first step towards refactoring to `update` command. We're going to use `named` library for this:

* https://hackage.haskell.org/package/named-0.2.0.0/docs/Named.html

ChShersh:
After looking at current template files I can propose the following separation (not sure whether it makes sense or not). @vrom911, please, look and leave any comments 🙂 

* `Template.Haskell`
    + `.cabal` file
    + library
    + prelude
    + executable
    + test-suite
    + bench
* `Template.Stack`
    + `stack.yaml`
* `Template.Doc`
    + `LICENSE`
    + `.gitignore`
    + `README.md`
    + `CHANGELOG.md`
    + `CONTRIBUTING.md`
    + `.stylish-haskell.yaml`
* `Template.CI`
    + `.travis.yml`
    + `appveyor.yml`

vrom911:
I'm not sure if `Template.Github` is better than `Template.CI`, but if that was me I would suggest it and also put `.gitignore` in there. Otherwise looks nice!

ChShersh:
@vrom911 Another possibility is to name it as `Template.Configs` and put there `.stylish-haskell.yaml` as well (but I don't like `Configs` name). Maybe something like `External` for _external tools_ or `Integrations` for GitHub + Travis + AppVeyor + stylish-haskell?

I also was thinking about having `Template.Cabal` for `.cabal` only file.

What do you think?

vrom911:
We can make `Cabal` module as well ( I have a feeling that in the future we will need to generate more cabal specific files, like `cabal.project`).

If you can come up with the intuitive name I don't mind having GitHub and stylish in one module, but for now I find `Github` module more pretty.

ChShersh:
@vrom911 I've updated the module hierarchy. Please, check, whether it makes more sense now 🙂

PR #129, opened 2018-08-21
[#125] Move Decision and GhcVer into their own modules

ChShersh:
This is only a step in breaking `Template` into smaller pieces. Breaking `Template` requires to introduce about 20 `newtype`s in `ProjectData` module. So I decided to move `GhcVer` and `Decision` into separate modules to not mix different things.

This PR also contains some other refactoring.

### DIFF
--- a/src/Summoner/Decision.hs
+++ b/src/Summoner/Decision.hs
@@ -1,3 +1,5 @@
+-- | Decision data type.
+
 module Summoner.Decision
        ( Decision (..)
        , decisionToBool
@@ -10,9 +12,6 @@ import Generics.Deriving.Semigroup (GSemigroup (..))
 
 import Summoner.Question (chooseYesNoBool, falseMessage, trueMessage)
 
-----------------------------------------------------------------------------
--- Decision data type
-----------------------------------------------------------------------------
 
 -- | Used for detecting the user decision during CLI input.
 data Decision = Idk | Nop | Yes
--- a/src/Summoner/Decision.hs
+++ b/src/Summoner/Decision.hs
@@ -1,0 +1,42 @@
+module Summoner.Decision
+       ( Decision (..)
+       , decisionToBool
+       ) where
+
+import Relude
+
+import Generics.Deriving.Monoid (GMonoid (..))
+import Generics.Deriving.Semigroup (GSemigroup (..))
+
+import Summoner.Question (chooseYesNoBool, falseMessage, trueMessage)
+
+----------------------------------------------------------------------------
+-- Decision data type
+----------------------------------------------------------------------------
+
+-- | Used for detecting the user decision during CLI input.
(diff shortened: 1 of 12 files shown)


src/Summoner/Decision.hs
```diff
@@ -0,0 +1,42 @@
...
+
+import Relude
+
+import Generics.Deriving.Monoid (GMonoid (..))
+import Generics.Deriving.Semigroup (GSemigroup (..))
+
+import Summoner.Question (chooseYesNoBool, falseMessage, trueMessage)
+
+----------------------------------------------------------------------------
+-- Decision data type
```
vrom911: I guess this can easily be removed or added as the module description.

src/Summoner/GhcVer.hs
```diff
@@ -0,0 +1,49 @@
...
+import Relude
+import Relude.Extra.Enum (inverseMap)
+
+-- | Represents some selected set of GHC versions.
+data GhcVer
+    = Ghc7103
+    | Ghc801
+    | Ghc802
+    | Ghc822
+    | Ghc843
```
vrom911: Should we already add 8.6.1? Or it's too soon?
It's okay with cabal, but stack will bring a lot of issues with it :(

src/Summoner/Template.hs
```diff
@@ -14,23 +14,23 @@ import Data.List (delete)
...
 
 ----------------------------------------------------------------------------
--- Stack File Creation
+-- Project Directory Creation
 ----------------------------------------------------------------------------
 
 memptyIfFalse :: Monoid m => Bool -> m -> m
 memptyIfFalse p val = if p then val else mempty
 
 -- | Creating template file to use in `stack new` command
```
vrom911: We are not using `stack new` anymore

src/Summoner/Question.hs
```diff
@@ -28,13 +29,25 @@ import System.FilePath ((</>))
...
 
 import qualified Data.Text as T
 import qualified Data.Text.IO as T
 import qualified Relude.Unsafe as Unsafe
 
+----------------------------------------------------------------------------
+-- Yes/No
+----------------------------------------------------------------------------
+
+data Answer = Y | N
```
vrom911: Perfect place, thanks!

src/Summoner/GhcVer.hs
ChShersh: Yeah, we don't have logic that allows to add `GhcVer` only to `cabal`. So I guess we need to wait for it...

src/Summoner/Template.hs
willbasky: Did we use it before?

vrom911: When it was `hs-init` a long time ago. But then we realised that it's very limited to what we actually need, and it was overcomplicated some stuff.

Issue #130, opened 2018-08-21
Show git revision version under `--version` command


ChShersh:
This article contains example:

* https://haskell-lang.org/library/optparse-applicative

And this is the used package (it's very light):

* https://hackage.haskell.org/package/gitrev-1.3.0/candidate/docs/Development-GitRev.html

We could have output like this:

```haskell
$ summon --version
Summoner v1.0.7 master@2ae047ba5e4a6f0f3e705a43615363ac006099c1 (Mon Jan 11 11:50:59 2016 +0800)
```

PR #132, opened 2018-08-21
[#130] Show git revision

willbasky:
Resolves [#130](https://github.com/kowainik/summoner/issues/130)

### DIFF
--- a/src/Summoner/Ansi.hs
+++ b/src/Summoner/Ansi.hs
@@ -4,11 +4,15 @@ module Summoner.Ansi
        ( Color (..)
        , putStrFlush
        , beautyPrint
+       , boldCode
+       , blueCode
        , bold
        , boldText
        , boldDefault
        , italic
+       , redCode
        , reset
+       , resetCode
        , prompt
        , setColor
        , successMessage
@@ -21,7 +25,7 @@ module Summoner.Ansi
 import Relude
 
 import System.Console.ANSI (Color (..), ColorIntensity (Vivid), ConsoleIntensity (BoldIntensity),
-                            ConsoleLayer (Foreground), SGR (..), setSGR)
+                            ConsoleLayer (Foreground), SGR (..), setSGR, setSGRCode)
 import System.IO (hFlush)
 
 ----------------------------------------------------------------------------
@@ -82,3 +86,9 @@ warningMessage = colorMessage Yellow
 successMessage = colorMessage Green
 infoMessage    = colorMessage Blue
 skipMessage    = colorMessage Cyan
+
+blueCode, boldCode, redCode, resetCode :: String
+redCode = setSGRCode [SetColor Foreground Vivid Red]
+blueCode = setSGRCode [SetColor Foreground Vivid Blue]
+boldCode = setSGRCode [SetConsoleIntensity BoldIntensity]
+resetCode = setSGRCode [Reset]
--- a/src/Summoner/CLI.hs
+++ b/src/Summoner/CLI.hs
@@ -18,13 +18,11 @@ import Options.Applicative (Parser, ParserInfo, command, execParser, flag, fullD
                             info, infoFooter, infoHeader, infoOption, long, metavar, optional,
                             progDesc, short, strArgument, strOption, subparser, switch)
 import Options.Applicative.Help.Chunk (stringChunk)
-import System.Console.ANSI (Color (..), ColorIntensity (..), ConsoleIntensity (..),
-                            ConsoleLayer (..), SGR (..), setSGRCode)
 import System.Directory (doesFileExist)
 
 import Paths_summoner (version)
-import Summoner.Ansi (Color (Green), beautyPrint, bold, errorMessage, infoMessage, setColor,
-                      warningMessage)
+import Summoner.Ansi (Color (Green), beautyPrint, blueCode, bold, boldCode, errorMessage,
+                      infoMessage, redCode, resetCode, setColor, warningMessage)
 import Summoner.Config (ConfigP (..), PartialConfig, defaultConfig, finalise, loadFileConfig)
 import Summoner.Decision (Decision (..))
 import Summoner.Default (defaultConfigFile, endLine)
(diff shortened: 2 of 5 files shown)


ChShersh:
@vrom911 @willbasky What if we can come up with something even more pretty to have extremely nice and cute interface? Something like this:

```shell
$ summon --version
Summoner v1.1.0
✧ Git revision: 1c742e1fb4f4beb6e92737dd9633b8af00a9e04a
✧ Commit date:  Tue Aug 21 14:26:13 2018 +0500
```

I like and idea with coloring. We can just use the same function similar to make directories bold in tree printing. What if we make strings `Summoner v1.1.0`, `Git revision:` and `Commit date:` and bold and also print them in some color (green or blue).

Also, 👍 for making `There are non-committed files.` part red.

src/Summoner/CLI.hs
```diff
@@ -116,14 +118,26 @@ versionP = infoOption summonerVersion
...
    <> short 'v'
    <> help "Show summoner's version"
+
+summonerVersion :: String
+summonerVersion = toString
+    [text|
+    $sVersion
+    $sHash
+    $sDate
+    $sDirty
```
ChShersh: It's better to not use `neat-interpolation` here to not have empty lines.

src/Summoner/CLI.hs
```diff
@@ -116,14 +118,26 @@ versionP = infoOption summonerVersion
...
   where
-    summonerVersion :: String
-    summonerVersion = concat
-        [ "Summoner v", showVersion version
-        , ", Git revision: ", $(gitHash)
-        , ", Commit date: ", $(gitCommitDate)
-        , if $(gitDirty) then ", there are non-committed files." else ""
-        ]
+    sVersion = toText $ setSGRCode [SetColor Foreground Vivid Blue]
+        <> "Summoner version: " <> setSGRCode [Reset] <>  showVersion version
```
ChShersh: I liked previous version for first line as `Summoner v1.1.0`

src/Summoner/CLI.hs
```diff
@@ -116,14 +118,26 @@ versionP = infoOption summonerVersion
...
+    |]
   where
-    summonerVersion :: String
-    summonerVersion = concat
-        [ "Summoner v", showVersion version
-        , ", Git revision: ", $(gitHash)
-        , ", Commit date: ", $(gitCommitDate)
-        , if $(gitDirty) then ", there are non-committed files." else ""
-        ]
+    sVersion = toText $ setSGRCode [SetColor Foreground Vivid Blue]
```
ChShersh: Let's also make colorful text bold

src/Summoner/CLI.hs
```diff
@@ -116,14 +118,26 @@ versionP = infoOption summonerVersion
...
+        <> "Summoner version: " <> setSGRCode [Reset] <>  showVersion version
+    sHash = toText $ setSGRCode [SetColor Foreground Vivid Blue]
+        <> "Git revision: " <> setSGRCode [Reset] <> $(gitHash)
+    sDate = toText $ setSGRCode [SetColor Foreground Vivid Blue]
+        <> "Commit date: " <> setSGRCode [Reset] <> $(gitCommitDate)
```
ChShersh: I would like to have `➤ ` prefix before `Git revision` and `Commit date:`

willbasky:
It looks like now:
default
is it worth adding bold?

src/Summoner/CLI.hs
```diff
@@ -120,20 +120,17 @@ versionP = infoOption summonerVersion
...
+        <> setSGRCode [SetConsoleIntensity BoldIntensity]
         <> "Summoner version: " <> setSGRCode [Reset] <>  showVersion version
     sHash = toText $ setSGRCode [SetColor Foreground Vivid Blue]
-        <> "Git revision: " <> setSGRCode [Reset] <> $(gitHash)
+        <> setSGRCode [SetColor Foreground Vivid Blue]
+        <> "➤ Git revision: " <> setSGRCode [Reset] <> $(gitHash)
     sDate = toText $ setSGRCode [SetColor Foreground Vivid Blue]
-        <> "Commit date: " <> setSGRCode [Reset] <> $(gitCommitDate)
+        <> setSGRCode [SetColor Foreground Vivid Blue]
+        <> "➤ Commit date: " <> setSGRCode [Reset] <> $(gitCommitDate)
```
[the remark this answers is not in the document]
willbasky: > But currently this module works only with IO.

Do you mean it doesn`t work for `pretty [blue, bold] "➤ Commit date: " <> $(gitCommitDate)` now?

src/Summoner/CLI.hs
```diff
@@ -122,18 +120,14 @@ versionP = infoOption summonerVersion
 summonerVersion :: String
 summonerVersion = toString $ unlines [sVersion, sHash, sDate, sDirty]
```
ChShersh: It seems better to write `toText` only once instead of adding `toText` before every variable:

```haskell
summonerVersion = toString $ unlines $ map toText [sVersion, ...]
```

src/Summoner/Ansi.hs
```diff
@@ -82,3 +86,9 @@ warningMessage = colorMessage Yellow
 successMessage = colorMessage Green
 infoMessage    = colorMessage Blue
 skipMessage    = colorMessage Cyan
+
+blueCode, boldCode, redCode, resetCode :: String
+redCode = setSGRCode [SetColor Foreground Vivid Red]
+blueCode = setSGRCode [SetColor Foreground Vivid Blue]
+boldCode = setSGRCode [SetConsoleIntensity BoldIntensity]
+resetCode = setSGRCode [Reset]
```
ChShersh: Very nice! This will work for now :+1:

willbasky:
`gitDirty` is always true on my side, even after all committed and pushed. :thinking:

ChShersh:
@willbasky Yeah, I agree with @vrom911. The fix is easy actually. Instead of having:

```haskell
[sVersion, sHash, sDate, sDirty]
```

You can use trick with list comprehensions and write something like this:

```haskell
[sVersion, sHash, sDate] ++ [sDirty | isDirty]
```

Instead of return empty string for `sDirty` you can return empty list if not dirty. We already use this trick to construct tree structure of the project:

https://github.com/kowainik/summoner/blob/5658df787f426a594d39539a1ce7ceb02ebeebad/src/Summoner/Template.hs#L54-L57

Issue #63, opened 2018-05-25
Add 'show ghc' and 'show license' commands


ChShersh:

ChShersh:
Note, that this is an extremely breaking change. It requires to change CLI interface. So we're going to introduce `new` command. And you can create packages with:

```
summon new name-of-the-package
```

Unfortunately, this is a required change to move further.

willbasky:
`summon show ghc` will show list of all available **ghc versions**? 
`summon show license MIT` to show **MIT** license text?
`summon show license BSD2` to show **BSD2** license text?

ChShersh:
@willbasky Yes, that's right. With one addition:

* `summon show license`: shows names of all supported licenses. So if now license name is given then it should print name of all available licenses with a hint like `Use show license [LICENSE_NAME]` command to see body of license.

willbasky:
> So we're going to introduce new command. And you can create packages with: `summon new name-of-the-package`

To make just interface blank without implementation in logic?

ChShersh:
> To make just interface blank without implementation in logic?

Sorry, if I wasn't clear. Currently, you can create package with the following command:

```
summon name-of-the-package
```

But after we add `show` command, we have conflict. Consider this usage example:

```
summon show
```

What this command should do? It's ambiguous because it has two possibilities:

1. Create package with name `show`.
2. Show `--help` message for `summon show` command.

Ambiguity in CLI is very bad. So there's only one option: change existing interface of `summoner`. So `summon name-of-the-package` will be no longer a valid command. To create package you need to write:

```
summon new name-of-the-package
```

PR #134, opened 2018-08-22
[#63] Add show commands

willbasky:
Resolves [#63](https://github.com/kowainik/summoner/issues/63)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
 * [#130](https://github.com/kowainik/summoner/issues/130):
   Show git revision version under `--version` command.
 * [#63](https://github.com/kowainik/summoner/issues/63):
-  Add `show --ghc` and `show --license` commands.
+  Add `show ghc`, `show licenses` and `show license <LICENSE_NAME>` commands.
 
 1.0.6
 =====
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@
   Print directories bold.
 * [#130](https://github.com/kowainik/summoner/issues/130):
   Show git revision version under `--version` command.
+* [#63](https://github.com/kowainik/summoner/issues/63):
+  Add `show --ghc` and `show --license` commands.
 
 1.0.6
 =====
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
 * [#130](https://github.com/kowainik/summoner/issues/130):
   Show git revision version under `--version` command.
 * [#63](https://github.com/kowainik/summoner/issues/63):
-  Add `show ghc`, `show licenses` and `show license <LICENSE_NAME>` commands.
+  Add `show ghc`, `show license` and `show license <LICENSE_NAME>` commands.
 
 1.0.6
 =====
--- a/README.md
+++ b/README.md
@@ -122,9 +122,15 @@ The other way to specify some particular `.toml` file is `summon new PROJECTNAME
 
 ##### CLI
 
-See the basic usage syntax below (you can check it out with `summon --help` command):
-
 ```
+Available commands:
+  new                      Create a new Haskell project
+  show                     Show available licenses or ghc versions
+
+---------------
+----- new -----
+---------------
+
 summon new PROJECT_NAME [--cabal] [--stack] [--ignore-config]
        [with [OPTIONS]] [without [OPTIONS]]
        [-f|--file FILENAME]  [--prelude-package PACKAGE_NAME]
@@ -148,10 +154,6 @@ Available global options:
 Available commands:
   with                     Specify options to enable
   without                  Specify options to disable
-  show                     Show command. (Support licenses, license and ghcs)
-  licenses                 Show available licenses
-  license LICENSE_NAME     Show specific license text
-  ghcs                     Show available ghc versions
 
 Available command options:
   -h, --help               Show this help text
@@ -165,6 +167,20 @@ Available command options:
   -t, --test               Tests
   -b, --benchmark          Benchmarks
 
+----------------
+----- show -----
+----------------
+
+summon show COMMAND [LICENSE_NAME]    
+  Show command. (Support licenses, license and ghcs)
+
+Available commands:
+  ghc                      Show available ghc versions
+  license                  Show available licenses
+  license [LICENSE_NAME]   Show specific license text
+
+Available global options:
+  -h, --help               Show this help text
 ```
 
 The options to be enabled/disabled can be specified while running the command.
--- a/README.md
+++ b/README.md
@@ -148,9 +148,13 @@ Available global options:
 Available commands:
   with                     Specify options to enable
   without                  Specify options to disable
+  show                     Show command. (Support licenses, license and ghcs)
+  licenses                 Show available licenses
+  license LICENSE_NAME     Show specific license text
+  ghcs                     Show available ghc versions
 
 Available command options:
-  -h,--help                Show this help text
+  -h, --help               Show this help text
   -g, --github             Github integration
   -p, --private            Create private GitHub repository
   -c, --travis             Travis CI integration
--- a/src/Summoner/CLI.hs
+++ b/src/Summoner/CLI.hs
@@ -12,29 +12,28 @@ module Summoner.CLI
 import Relude
 import Relude.Extra.Enum (universe)
 
-import Data.Aeson (decodeStrict)
-import Data.ByteString.Char8 (pack)
 import Data.Version (showVersion)
 import Development.GitRev (gitCommitDate, gitDirty, gitHash)
 import NeatInterpolation (text)
-import Options.Applicative (Parser, ParserInfo, command, execParser, flag, flag', fullDesc, help,
-                            helper, info, infoFooter, infoHeader, infoOption, long, metavar,
-                            optional, progDesc, short, strArgument, strOption, subparser, switch)
+import Options.Applicative (Parser, ParserInfo, command, execParser, flag, fullDesc, help, helper,
+                            info, infoFooter, infoHeader, infoOption, long, metavar, optional,
+                            progDesc, short, strArgument, strOption, subparser, switch)
 import Options.Applicative.Help.Chunk (stringChunk)
 import System.Directory (doesFileExist)
-import System.Process (readProcess)
 
 import Paths_summoner (version)
 import Summoner.Ansi (Color (Green), beautyPrint, blueCode, bold, boldCode, errorMessage,
-                      infoMessage, redCode, resetCode, setColor, warningMessage)
+                      infoMessage, redCode, resetCode, setColor, successMessage, warningMessage)
 import Summoner.Config (ConfigP (..), PartialConfig, defaultConfig, finalise, loadFileConfig)
 import Summoner.Decision (Decision (..))
 import Summoner.Default (defaultConfigFile, endLine)
 import Summoner.GhcVer (showGhcVer)
-import Summoner.License (License (..), LicenseName (..), githubLicenseQueryNames, parseLicenseName)
+import Summoner.License (License (..), LicenseName (..), getLicense, parseLicenseName)
 import Summoner.Project (generateProject)
 import Summoner.ProjectData (CustomPrelude (..))
 import Summoner.Validation (Validation (..))
+
+import qualified Data.Text as T
 
 ---------------------------------------------------------------------------
 -- CLI
@@ -53,29 +52,25 @@ runShow :: ShowOpts -> IO ()
 runShow opts =
     case opts of
         -- show list of all available GHC versions
-        ShowGhc          -> putText $ unlines $ map showGhcVer universe
-        ShowLicense name -> case name of
-            -- show a list of all available licenses
-            LicenseList Nothing  -> putText $ unlines $ map (\x -> show (x :: LicenseName)) universe
+        GhcList -> ghcVers
+        -- show a list of all available licenses
+        LicenseList -> licenses
             -- show a specific license
-            LicenseList (Just n) -> do
-                -- check a user`s input
-                license <- whenNothing (parseLicenseName (toText n))
-                    (fail "This wasn't a valid choice.")
-                -- get a link to github
-                let licenseLink = "https://api.github.com/licenses/" <> githubLicenseQueryNames license
-                -- get a JSON of specific license from github
-                licenseJson <-
-                    readProcess "curl"
-                                [ toString licenseLink
-                                , "-H"
-                                , "Accept: application/vnd.github.drax-preview+json"
-                                ]
-                                ""
-                -- get a license`s text and show
-                case (decodeStrict $ pack licenseJson) :: Maybe License of
-                    Just t  -> putStr $ unLicense t
-                    Nothing -> error "Broken predefined license list"
+        LicenseText name ->
+            case parseLicenseName (toText name) of
+                Nothing -> do
+                    errorMessage "This wasn't a valid choice."
+                    infoMessage "Here is the list of supported licenses"
+                    licenses
+                Just licenseName -> do
+                    -- get and show a license`s text
+                    mLicenseText <- getLicense licenseName
+                    case mLicenseText :: Maybe License of
+                        Just t  -> putStr $ unLicense t
+                        Nothing -> error "Broken predefined license list"
+  where
+    ghcVers = mapM_ (successMessage . T.append "\10148 " . showGhcVer) (reverse universe)
+    licenses = mapM_ (successMessage . (\ x -> T.append "\10148 " $ show (x :: LicenseName))) universe
 
 runNew :: NewOpts -> IO ()
 runNew NewOpts{..} = do
@@ -136,9 +131,8 @@ data NewOpts = NewOpts
     , cliConfig   :: PartialConfig  -- ^ config gathered during CLI
     }
 
-data ShowOpts = ShowGhc | ShowLicense LicenseList
-
-newtype LicenseList = LicenseList (Maybe String)
+-- | Commands parsed with @show@ command
+data ShowOpts = GhcList | LicenseList | LicenseText String
 
 ----------------------------------------------------------------------------
 -- Parsers
@@ -178,21 +172,14 @@ summonerP = subparser
 
 -- | Parses options of the @show@ command.
 showP :: Parser Command
-showP = ShowInfo <$> (ghcP <|> ShowLicense <$> licenseP)
+showP = ShowInfo <$> subparser
+    ( command "ghcs" (info (helper <*> pure GhcList) $ progDesc "Show available ghc versions")
+   <> command "licenses" (info (helper <*> pure LicenseList) $ progDesc "Show available licenses")
+   <> command "license" (info (helper <*> licenseText) $ progDesc "Show specific license text")
+    )
 
-ghcP :: Parser ShowOpts
-ghcP = flag' ShowGhc $ long "ghc" <> short 'g' <> help "Show available ghc versions"
-
-licenseP :: Parser LicenseList
-licenseP = subparser $
-    command "licenses" (info (helper <*> licenseList) $ progDesc "Show available licenses")
-
-licenseList :: Parser LicenseList
-licenseList =  LicenseList <$> optional licenseText
-
-licenseText :: Parser String
-licenseText = strOption $
-    long "name" <> short 'n' <> metavar "LICENSE_NAME" <> help "Show license text"
+licenseText :: Parser ShowOpts
+licenseText = LicenseText <$> strArgument (metavar "LICENSE_NAME")
 
 -- | Parses options of the @new@ command.
 newP :: Parser Command
@@ -334,9 +321,6 @@ stackP :: Parser Decision
 stackP = flag Idk Yes
        $ long "stack"
       <> help "Stack support for the project"
-
--- showGhc :: Parser Command
--- showGhc = subparser Info $
 
 ----------------------------------------------------------------------------
 -- Beauty util
--- a/src/Summoner/CLI.hs
+++ b/src/Summoner/CLI.hs
@@ -1,8 +1,8 @@
-{-# LANGUAGE ApplicativeDo       #-}
-{-# LANGUAGE QuasiQuotes         #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TemplateHaskell     #-}
-{-# LANGUAGE TupleSections       #-}
+{-# LANGUAGE ApplicativeDo    #-}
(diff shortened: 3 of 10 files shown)


src/Summoner/CLI.hs
```diff
@@ -41,6 +47,35 @@ summon = execParser prsr >>= runCommand
 runCommand :: Command -> IO ()
 runCommand = \case
     New opts -> runNew opts
+    ShowInfo opts -> runShow opts
+
+runShow :: ShowOpts -> IO ()
+runShow opts =
+    case opts of
+        -- show list of all available GHC versions
+        ShowGhc          -> putText $ unlines $ map showGhcVer universe
```
vrom911: I guess we can improve the outlook a bit. As it's a list we can add the arrow symbol (the one we used in git revision).
And also it would be nice to shoe it in reversed order (from latest to oldest).

src/Summoner/CLI.hs
```diff
@@ -41,6 +47,35 @@ summon = execParser prsr >>= runCommand
...
+        ShowLicense name -> case name of
+            -- show a list of all available licenses
+            LicenseList Nothing  -> putText $ unlines $ map (\x -> show (x :: LicenseName)) universe
+            -- show a specific license
+            LicenseList (Just n) -> do
+                -- check a user`s input
+                license <- whenNothing (parseLicenseName (toText n))
+                    (fail "This wasn't a valid choice.")
```
vrom911: I guess in this case all that we need to is to `errorMessage`(so it will be red) this but also on the next line print the list of licences we support saying something like: "Here is the list of supported licenses".

src/Summoner/CLI.hs
```diff
@@ -41,6 +47,35 @@ summon = execParser prsr >>= runCommand
...
+                    readProcess "curl"
+                                [ toString licenseLink
+                                , "-H"
+                                , "Accept: application/vnd.github.drax-preview+json"
+                                ]
+                                ""
+                -- get a license`s text and show
+                case (decodeStrict $ pack licenseJson) :: Maybe License of
+                    Just t  -> putStr $ unLicense t
+                    Nothing -> error "Broken predefined license list"
```
vrom911: This chunk of code is actually absolutely the same as the one user in `Project` module. So it makes sense to put it into `License` module and use in both cases.

src/Summoner/CLI.hs
```diff
@@ -99,6 +136,10 @@ data NewOpts = NewOpts
     , cliConfig   :: PartialConfig  -- ^ config gathered during CLI
     }
 
+data ShowOpts = ShowGhc | ShowLicense LicenseList
```
vrom911: It would be nice to add documentation to this data type and to both it constructors.

src/Summoner/CLI.hs
```diff
@@ -99,6 +136,10 @@ data NewOpts = NewOpts
...
+
+newtype LicenseList = LicenseList (Maybe String)
```
vrom911: and to this newtype as well

src/Summoner/CLI.hs
```diff
@@ -128,14 +169,31 @@ summonerVersion = toString $ intercalate "\n" $ [sVersion, sHash, sDate] ++ [sDi
...
+    $ command "new" (info (helper <*> newP) $ progDesc "Create a new Haskell project")
+   <> command "show" (info (helper <*> showP) $ progDesc "Show available licenses or ghc versions")
 
 ----------------------------------------------------------------------------
 -- New command parsers
 ----------------------------------------------------------------------------
 
+-- | Parses options of the @show@ command.
+showP :: Parser Command
+showP = ShowInfo <$> (ghcP <|> ShowLicense <$> licenseP)
```
vrom911: I think we should be consistent in the choice of the command/flag.
Either both of the `ghc` and `licence` should be commands, or both of them flags.
(I guess the command is preferable)

src/Summoner/CLI.hs
```diff
@@ -128,14 +169,31 @@ summonerVersion = toString $ intercalate "\n" $ [sVersion, sHash, sDate] ++ [sDi
...
+licenseP :: Parser LicenseList
+licenseP = subparser $
+    command "licenses" (info (helper <*> licenseList) $ progDesc "Show available licenses")
+
+licenseList :: Parser LicenseList
+licenseList =  LicenseList <$> optional licenseText
+
+licenseText :: Parser String
+licenseText = strOption $
+    long "name" <> short 'n' <> metavar "LICENSE_NAME" <> help "Show license text"
```
vrom911: I guess there should be a way to avoid this `name` option, so we can:
```
summon show --licence
# and
summon show --licence MIT
```

src/Summoner/CLI.hs
```diff
@@ -277,6 +335,9 @@ stackP = flag Idk Yes
        $ long "stack"
       <> help "Stack support for the project"
 
+-- showGhc :: Parser Command
+-- showGhc = subparser Info $
+
```
vrom911: 🤔

src/Summoner/CLI.hs
ChShersh: I would like to have `stack`-like interface for this. For example, in `stack` there's `ls` command that shows different things. And I can do:

```shell
$ stack ls # prints error that command is missing
$ stack ls dependencies # shows dependencies
$ stack ls snapshots # shows snapshots
```

It would be great if we can write something like that. Specifically, I propose to have the following interface:

```shell
$ summon show # error
$ summon show ghc
$ summon show licenses
$ summon show license MIT
```
Also, what do you think about renaming `show` to `ls`?

willbasky: > $ summon show license MIT
> summon show --licence MIT

I am not sure is it possible to have `optional metavar`. Task is to have double functionality. 
First `summon show license` shows list of licenses. And second `summon show license MIT` shows particular license`s text.
And I am not sure is `command` like `summon show license MIT` has **metavar** possibility or not.

ChShersh: @willbasky It's okay to have two different words for different commands:

* `licenses` to show list of all supported licences
* `license` to show only specific license

Also, I'm thinking about renaming `show ghc` to `show ghcs` to make naming more consistent.

willbasky: About `ls` or `show`. We need data what users prefers or some arguments why one is better then other one.

ChShersh: `ls` is shorter, but usually `ls` is used to display multiple things (like entries in directory), so it might be confusing to have `summon ls license MIT`
`show` is more human readable, but longer. Though, only 2 characters longer.

src/Summoner/CLI.hs
vrom911: I guess it's better to pattern match here and put all the following code under the `Just` case. Because you shouldn't give any value to the license if you can not parse it. But also we don't want to see any errors. Just print the `errorMessage` and the text I mentioned above, without failing.

README.md
```diff
@@ -148,9 +148,13 @@ Available global options:
 Available commands:
   with                     Specify options to enable
   without                  Specify options to disable
+  show                     Show command. (Support licenses, license and ghcs)
```
vrom911: This is not completely correct documentation. Now as we have not only `new` command we need to reorganize this section. For this `--help` command will be very helpful 😉 

So the command section should contain the only commands we have now: `new`, `show`. (Use `summon --help` to see this).
Then for each of this commands you can write what we have at the moment. (Use `summon COMMAND --help` to see).

src/Summoner/CLI.hs
```diff
@@ -41,6 +46,31 @@ summon = execParser prsr >>= runCommand
...
+        LicenseText name ->
+            case parseLicenseName (toText name) of
+                Nothing -> do
+                    errorMessage "This wasn't a valid choice."
+                    infoMessage "Here is the list of supported licenses"
+                    licenses
+                Just licenseName -> do
+                    -- get and show a license`s text
+                    mLicenseText <- getLicense licenseName
+                    case mLicenseText :: Maybe License of
```
vrom911: This function is also shared with the `Project` module, you can put it into the separate function as well 🙂

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@
   Print directories bold.
 * [#130](https://github.com/kowainik/summoner/issues/130):
   Show git revision version under `--version` command.
+* [#63](https://github.com/kowainik/summoner/issues/63):
+  Add `show ghc`, `show licenses` and `show license <LICENSE_NAME>` commands.
```
ChShersh: I know think that it's better to not have `licenses` as a separate command. It would be even better if we could have less commands. And support:

```shell
$ summon show ghc
$ summon show license
$ summon show license MIT
```

This gives us open design space for putting something under the following command:

```shell
$ summon show ghc 8.4.3
```

src/Summoner/CLI.hs
```diff
@@ -41,6 +46,31 @@ summon = execParser prsr >>= runCommand
 runCommand :: Command -> IO ()
 runCommand = \case
     New opts -> runNew opts
+    ShowInfo opts -> runShow opts
+
+runShow :: ShowOpts -> IO ()
+runShow opts =
+    case opts of
+        -- show list of all available GHC versions
+        GhcList -> ghcVers
```
ChShersh: It's better to rename `ghcVers` to `showGhcVers`.

src/Summoner/CLI.hs
```diff
@@ -41,6 +46,31 @@ summon = execParser prsr >>= runCommand
 runCommand :: Command -> IO ()
 runCommand = \case
     New opts -> runNew opts
+    ShowInfo opts -> runShow opts
+
+runShow :: ShowOpts -> IO ()
+runShow opts =
```
ChShersh: I think this function can be written shorter with `-XLambdaCase` extension

src/Summoner/CLI.hs
```diff
@@ -41,6 +46,31 @@ summon = execParser prsr >>= runCommand
...
+    case opts of
+        -- show list of all available GHC versions
+        GhcList -> ghcVers
+        -- show a list of all available licenses
+        LicenseList -> licenses
```
ChShersh: Same: `showLicenses`.

src/Summoner/CLI.hs
```diff
@@ -41,6 +46,31 @@ summon = execParser prsr >>= runCommand
...
+                    infoMessage "Here is the list of supported licenses"
+                    licenses
+                Just licenseName -> do
+                    -- get and show a license`s text
+                    mLicenseText <- getLicense licenseName
+                    case mLicenseText :: Maybe License of
+                        Just t  -> putStr $ unLicense t
+                        Nothing -> error "Broken predefined license list"
+  where
+    ghcVers = mapM_ (successMessage . T.append "\10148 " . showGhcVer) (reverse universe)
```
ChShersh: This is strange. Why not just add unicode arrow like it's done in `version`?

ChShersh: Also, I think it's better to use blue color here instead of green. The convention is that green for success and blue is for information.

src/Summoner/CLI.hs
```diff
@@ -41,6 +46,31 @@ summon = execParser prsr >>= runCommand
...
+    licenses = mapM_ (successMessage . (\ x -> T.append "\10148 " $ show (x :: LicenseName))) universe
```
ChShersh: It's always better to write explicit type of functions inside `where` instead of specifying types of variables locally like `x :: LicenseName`.

src/Summoner/License.hs
```diff
@@ -76,3 +79,10 @@ customizeLicense l t nm year
             (beforeN, withN) = T.span (/= '[') afterY
             afterN           = T.tail $ T.dropWhile (/= ']') withN
         in  beforeY <> year <> beforeN <> nm <> afterN
+
+getLicense :: LicenseName -> IO (Maybe License)
+getLicense name = do
```
ChShersh: It's better to name this function as `fetchLicense` or `downloadLicense`. `get*` prefix is too generic.

src/Summoner/CLI.hs
willbasky: `licenses`  has type `IO ()`. It doesn`t explicit what is inside. Do you mean something another?

vrom911: Even if it's `IO` still better to write everything explicitly. We even have a note on it in the style guide.

src/Summoner/License.hs
```diff
@@ -80,9 +80,12 @@ customizeLicense l t nm year
             afterN           = T.tail $ T.dropWhile (/= ']') withN
         in  beforeY <> year <> beforeN <> nm <> afterN
 
-fetchLicense :: LicenseName -> IO (Maybe License)
+fetchLicense :: LicenseName -> IO Text
```
vrom911: What is the point of returning `Text`? We shouldn't sacrifice type safety here without any reasons.

willbasky:
```
let licenseText = case decodeStrict $ pack licenseJson of
            Just t  -> unLicense t
            Nothing -> error "Broken predefined license list"
    pure licenseText
```
Do you mean to return not `licenseText :: Text` but `t :: License`?

src/Summoner/CLI.hs
willbasky: > Even if it's IO still better to write everything explicitly.

:+1: 

> We even have a note on it in the style guide.

Could you show me the style guide? I couldn`t find the note here https://github.com/kowainik/org

vrom911: Yes, sure. Currently, we are following [this style guide](https://github.com/serokell/serokell-util/blob/master/serokell-style.md#serokell-haskell-style-guide).

Sorry, the link in our `org` is broken :disappointed:, I will fix it!

src/Summoner/Project.hs
```diff
@@ -45,7 +45,7 @@ generateProject projectName Config{..} = do
     -- License creation
     fetchedLicense <- fetchLicense license
     year <- currentYear
-    let licenseText = customizeLicense license fetchedLicense nm year
+    let licenseText = customizeLicense license (unLicense fetchedLicense) nm year
```
vrom911: Hmm, I guess it's good to make `customizeLicense` receive `License` as well and unwrap it only there. We have three `Text` types here otherwise, which is not very good. Could you fix this as well? :slightly_smiling_face:

src/Summoner/CLI.hs
```diff
@@ -41,6 +46,30 @@ summon = execParser prsr >>= runCommand
...
+                    errorMessage "This wasn't a valid choice."
+                    infoMessage "Here is the list of supported licenses:"
+                    showLicenses
+                    -- get and show a license`s text
+                Just licenseName -> do
+                    fetchedLicense <- fetchLicense licenseName
+                    putTextLn $ unLicense fetchedLicense
+  where
+    showGhcVers :: IO ()
+    showGhcVers = mapM_ (infoMessage . T.append "➤ " . showGhcVer) (reverse universe)
```
ChShersh: I like `reverse universe` :smirk:

src/Summoner/Project.hs
```diff
@@ -46,18 +43,9 @@ generateProject projectName Config{..} = do
     license  <- choose parseLicenseName "License: " $ ordNub (cLicense : universe)
```
ChShersh: I think that this variable to the left of `<-` should be called `licenseName`

src/Summoner/Project.hs
```diff
@@ -46,18 +43,9 @@ generateProject projectName Config{..} = do
...
     -- License creation
-    let licenseLink = "https://api.github.com/licenses/" <> githubLicenseQueryNames license
-    licenseJson <-
-      readProcess "curl"
-                  [ toString licenseLink
-                  , "-H"
-                  , "Accept: application/vnd.github.drax-preview+json"
-                  ]
-                  ""
+    fetchedLicense <- fetchLicense license
```
ChShersh: And here you can just write: `license <- fetchLicense licenseName`

src/Summoner/Project.hs
```diff
@@ -46,18 +43,9 @@ generateProject projectName Config{..} = do
...
     year <- currentYear
-    let licenseText = case (decodeStrict $ pack licenseJson) :: Maybe License of
-            Just t  -> customizeLicense license (unLicense t) nm year
-            Nothing -> error "Broken predefined license list"
+    let licenseText = customizeLicense license (unLicense fetchedLicense) nm year
```
ChShersh: :+1: on @vrom911 suggestion. I would really appreciate that `customizeLicense` function takes `License` and returns `License` as well.

src/Summoner/CLI.hs
```diff
@@ -1,7 +1,8 @@
-{-# LANGUAGE ApplicativeDo   #-}
-{-# LANGUAGE QuasiQuotes     #-}
-{-# LANGUAGE TemplateHaskell #-}
-{-# LANGUAGE TupleSections   #-}
+{-# LANGUAGE ApplicativeDo       #-}
+{-# LANGUAGE QuasiQuotes         #-}
+{-# LANGUAGE ScopedTypeVariables #-}
```
vrom911: Can you add `ScopedTypeVariables` to the default extensions in the `.cabal` file. And also `TypeApplications`. It's very strange we don't have it in there..

src/Summoner/CLI.hs
```diff
@@ -66,10 +67,8 @@ runShow = \case
                     fetchedLicense <- fetchLicense licenseName
                     putTextLn $ unLicense fetchedLicense
   where
-    showGhcVers :: IO ()
-    showGhcVers = mapM_ (infoMessage . T.append "➤ " . showGhcVer) (reverse universe)
-    showLicenses :: IO ()
-    showLicenses = mapM_ (infoMessage . (\ x -> T.append "➤ " $ show (x :: LicenseName))) universe
+    showGhcVers :: IO () = mapM_ (infoMessage . T.append "➤ " . showGhcVer) (reverse universe)
```
vrom911: Better to come back to the previous version with the types on the separate lines 🙂

src/Summoner/License.hs
```diff
@@ -68,13 +68,14 @@ githubLicenseQueryNames = \case
 parseLicenseName :: Text -> Maybe LicenseName
 parseLicenseName = inverseMap show
 
-customizeLicense :: LicenseName -> Text -> Text -> Text -> Text
+customizeLicense :: LicenseName -> License -> Text -> Text -> License
 customizeLicense l t nm year
```
vrom911: instead of `t` you can write `licence@(License licenseText)`, so no need to add another function in `where` clause

CHANGELOG.md
ChShersh: Here should be `show license` instead of `show licenses`

src/Summoner/CLI.hs
```diff
@@ -41,6 +47,28 @@ summon = execParser prsr >>= runCommand
...
+                    errorMessage "This wasn't a valid choice."
+                    infoMessage "Here is the list of supported licenses:"
+                    showLicenses
+                    -- get and show a license`s text
+                Just licenseName -> do
+                    fetchedLicense <- fetchLicense licenseName
+                    putTextLn $ unLicense fetchedLicense
+  where
+    showGhcVers :: IO () = mapM_ (infoMessage . T.append "➤ " . showGhcVer) (reverse universe)
+    showLicenses :: IO () = mapM_ (infoMessage . (\ x -> T.append "➤ " $ show (x :: LicenseName))) universe
```
ChShersh: Actually, I see could duplication here. You write the same function twice:

```haskell
showBulletList :: (Show a) => [a] -> IO ()
```

So it's better to remove code duplication and have `showEnumList` function only. And then you can call this function like this:

```haskell
showBulletList @GhcVer (reverse universe)
showBulletList @LicenseName universe
```

src/Summoner/CLI.hs
```diff
@@ -128,14 +161,23 @@ summonerVersion = toString $ intercalate "\n" $ [sVersion, sHash, sDate] ++ [sDi
 -- All possible commands.
 summonerP :: Parser Command
 summonerP = subparser
-    ( command "new"
-      (info (helper <*> newP) $ progDesc "Create a new Haskell project")
-    )
+    $ command "new" (info (helper <*> newP) $ progDesc "Create a new Haskell project")
+   <> command "show" (info (helper <*> showP) $ progDesc "Show available licenses or ghc versions")
```
ChShersh: Better to write `supported` instead of `available`

src/Summoner/CLI.hs
```diff
@@ -128,14 +161,23 @@ summonerVersion = toString $ intercalate "\n" $ [sVersion, sHash, sDate] ++ [sDi
...
 
 ----------------------------------------------------------------------------
 -- New command parsers
 ----------------------------------------------------------------------------
 
+-- | Parses options of the @show@ command.
+showP :: Parser Command
+showP = ShowInfo <$> subparser
+    ( command "ghc" (info (helper <*> pure GhcList) $ progDesc "Show available ghc versions")
```
ChShersh: `supported` here as well

src/Summoner/Template.hs
```diff
@@ -57,7 +58,7 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
  ++ [File "b" scriptSh | script]
   where
     -- Creates module name from the name of the project
```
ChShersh: Let's move this comment before `libModuleName` function

src/Summoner/CLI.hs
willbasky: Why is it better?

src/Summoner/CLI.hs
willbasky: I added `TypeApplications` to `.cabal` but it didn`t help.

src/Summoner/CLI.hs
```diff
@@ -1,7 +1,8 @@
-{-# LANGUAGE ApplicativeDo   #-}
-{-# LANGUAGE QuasiQuotes     #-}
-{-# LANGUAGE TemplateHaskell #-}
-{-# LANGUAGE TupleSections   #-}
+{-# LANGUAGE ApplicativeDo    #-}
+{-# LANGUAGE QuasiQuotes      #-}
+{-# LANGUAGE TemplateHaskell  #-}
+{-# LANGUAGE TupleSections    #-}
+{-# LANGUAGE TypeApplications #-}
```
ChShersh: You can remove `TypeApplications` from this module since it's now in `default-extensions`

willbasky: If it is removed, parser error returns in lines where it uses.

vrom911: Can you show us which error you are facing?

willbasky: default

vrom911: Well, it's not the compilation error. It's `hlint` error. You need to modify your `hlint` file to say there about this extension. It shouldn't affect the library code.

Issue #122, opened 2018-08-10
Even if github integration is turned off, the changelog file contains links to github


ShrykeWindgrace:
In a new project CHANGELOG.md contains
```
Change log
==========

ProjectName uses [PVP Versioning][1].
The change log is available [on GitHub][2].

0.0.0
=====
* Initially created.

[1]: https://pvp.haskell.org
[2]: https://github.com/RepositoryOwner/ProjectName/releases
```
if when I did
```
Add GitHub integration? [y]/n
  ->   n
```
during setup.

```
summon --version
1.0.6
```

ShrykeWindgrace:
That great, thank you=)

I guess the licence badge in README.md should also be affected by this toggle:
```
[![MIT license](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/RepositoryOwner/ProjectName/blob/master/LICENSE)
```

PR #136, opened 2018-08-24
[#122] Fix changelog and readme

willbasky:
Resolves [#122](https://github.com/kowainik/summoner/issues/122)

### DIFF
--- a/src/Summoner/Template.hs
+++ b/src/Summoner/Template.hs
@@ -298,14 +298,17 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
 
     -- create README template
     readme :: Text
-    readme = T.intercalate "\n" $ filter (/="") 
-        [ ("# " <> repo <> "\n")
+    readme = T.intercalate "\n" $
+        [ "# " <> repo
+        , ""
         , hackage
         , licenseBadge
-        , stackBadges
-        , travisBadge
-        , appVeyorBadge
-        , ("\n" <> description)
+        ]
+     ++ [stackBadges | stack]
+     ++ [travisBadge | travis]
+     ++ [appVeyorBadge | appVey]
+     ++ [""
+        , description
         ]
       where
         hackageShield :: Text =
@@ -316,12 +319,8 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
 
         licenseShield :: Text =
             "https://img.shields.io/badge/license-" <> T.replace "-" "--" license <> "-blue.svg"
-        licenseLink :: Text = memptyIfFalse github $ 
-            "https://github.com/" <> owner <> "/" <> repo <> "/blob/master/LICENSE"
-        licenseUnLink :: Text = memptyIfFalse (not github) $ "LICENSE"
-
         licenseBadge :: Text = 
-            makeBadge (license <> " license") licenseShield (licenseLink <> licenseUnLink)
+            makeBadge (license <> " license") licenseShield "LICENSE"
 
         stackShieldLts :: Text =
             "http://stackage.org/package/" <> repo <> "/badge/lts"
@@ -332,23 +331,23 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
             "http://stackage.org/package/" <> repo <> "/badge/nightly"
         stackLinkNightly :: Text =
             "http://stackage.org/nightly/package/" <> repo
-        stackBadges :: Text = T.intercalate "\n" 
-            $ [makeBadge "Stackage Lts" stackShieldLts stackLinkLts | stack]
-           ++ [makeBadge "Stackage Nightly" stackShieldNightly stackLinkNightly | stack]
+        stackBadges :: Text 
+            = makeBadge "Stackage Lts" stackShieldLts stackLinkLts
+           <> makeBadge "Stackage Nightly" stackShieldNightly stackLinkNightly
             
         travisShield :: Text =
             "https://secure.travis-ci.org/" <> owner <> "/" <> repo <> ".svg"
         travisLink :: Text =
             "https://travis-ci.org/" <> owner <> "/" <> repo
-        travisBadge :: Text = T.concat 
-            [makeBadge "Build status" travisShield travisLink | travis]
+        travisBadge :: Text = 
+            makeBadge "Build status" travisShield travisLink
 
         appVeyorShield :: Text =
             "https://ci.appveyor.com/api/projects/status/github/" <> owner <> "/" <> repo <> "?branch=master&svg=true"
         appVeyorLink :: Text =
             "https://ci.appveyor.com/project/" <> owner <> "/" <> repo
-        appVeyorBadge :: Text = T.concat 
-            [makeBadge "Windows build status" appVeyorShield appVeyorLink | appVey]
+        appVeyorBadge :: Text = 
+            makeBadge "Windows build status" appVeyorShield appVeyorLink
 
         makeBadge :: Text -> Text -> Text -> Text
         makeBadge title shield link = "[![" <> title <> "](" <> shield <> ")](" <> link <> ")"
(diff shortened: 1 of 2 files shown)


src/Summoner/Template.hs
```diff
@@ -298,24 +298,33 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
...
 
-        [![Hackage]($hackageShield)]($hackageLink)
-        [![$license license](${licenseShield})](${licenseLink})
-        $stackBadges
-        $travisBadge
-        $appVeyorBadge
-
+        $hackage
+        $licenseBadge
+        $maybeSomeBadges
```
ChShersh: The idea is to not use `neat-interpolation` at all for `readme`. This file is somewhat special. And since we have a lot of conditionals here, it's better to use combination of `intercalate` function with list comprehension to not have empty lines at all but still have nice structure for the `README.md` file

src/Summoner/Template.hs
```diff
@@ -298,24 +298,33 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
 
     -- create README template
     readme :: Text
-    readme =
+    readme =
```
ChShersh: :scream: Extra space :scream:

src/Summoner/Template.hs
```diff
@@ -326,31 +335,28 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
...
             |]
 
         travisShield :: Text =
-          "https://secure.travis-ci.org/" <> owner <> "/" <> repo <> ".svg"
+            "https://secure.travis-ci.org/" <> owner <> "/" <> repo <> ".svg"
         travisLink :: Text =
-          "https://travis-ci.org/" <> owner <> "/" <> repo
-        travisBadge :: Text = memptyIfFalse travis
+            "https://travis-ci.org/" <> owner <> "/" <> repo
+        travisBadge :: Text =
```
ChShersh: I see that we have a lot of badges. And since we're not going to use `neat-interpolation` for `README.md` file anymore, I think it makes sense to introduce function like `makeBadge` so we can just use 

```haskell
mkBadge "Stackage Lts" stackShieldLts stackLinkLts
```

And same for other badges. Though, you still can use `neat-interpolation` to implement `mkBadge` function itself.

src/Summoner/Template.hs
```diff
@@ -298,24 +298,30 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
...
-        |]
+    readme = T.intercalate "\n" $ filter (/="") 
+        [ ("# " <> repo <> "\n")
+        , hackage
+        , licenseBadge
+        , stackBadges
+        , travisBadge
+        , appVeyorBadge
+        , ("\n" <> description)
+        ]
```
vrom911: instead of `filter (/="")` and memptyIfFalse in each component you can do
```haskell
    [ "# " <> repo
    , ""
    ] 
 ++ [ "some one line text | conditionBool ]
...
```

src/Summoner/Template.hs
```diff
@@ -298,24 +298,30 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
...
+            "https://img.shields.io/hackage/v/" <> repo <> ".svg"
         hackageLink :: Text =
-          "https://hackage.haskell.org/package/" <> repo
+            "https://hackage.haskell.org/package/" <> repo
+        hackage :: Text = makeBadge "Hackage" hackageShield hackageLink
+
+        licenseShield :: Text =
+            "https://img.shields.io/badge/license-" <> T.replace "-" "--" license <> "-blue.svg"
+        licenseLink :: Text = memptyIfFalse github $ 
+            "https://github.com/" <> owner <> "/" <> repo <> "/blob/master/LICENSE"
```
vrom911: This is not needed. The license link should always be the one that you wrote in the next line.

src/Summoner/Template.hs
```diff
@@ -419,17 +420,21 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
...
+        $githubLine
 
         0.0.0
         =====
         * Initially created.
 
         [1]: https://pvp.haskell.org
-        [2]: https://github.com/${owner}/${repo}/releases
+        $githubFootNote
         $endLine
```
vrom911: This can be removed.

src/Summoner/Template.hs
```diff
@@ -332,23 +331,23 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
             "http://stackage.org/package/" <> repo <> "/badge/nightly"
         stackLinkNightly :: Text =
             "http://stackage.org/nightly/package/" <> repo
-        stackBadges :: Text = T.intercalate "\n" 
-            $ [makeBadge "Stackage Lts" stackShieldLts stackLinkLts | stack]
-           ++ [makeBadge "Stackage Nightly" stackShieldNightly stackLinkNightly | stack]
+        stackBadges :: Text 
+            = makeBadge "Stackage Lts" stackShieldLts stackLinkLts
+           <> makeBadge "Stackage Nightly" stackShieldNightly stackLinkNightly
```
ChShersh: There should be empty line between these two badges

src/Summoner/Template.hs
```diff
@@ -298,14 +298,17 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
 
     -- create README template
     readme :: Text
-    readme = T.intercalate "\n" $ filter (/="") 
-        [ ("# " <> repo <> "\n")
+    readme = T.intercalate "\n" $
+        [ "# " <> repo
```
ChShersh: This README structure looks nice now!

src/Summoner/Template.hs
```diff
@@ -326,31 +331,26 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
...
+            "https://ci.appveyor.com/project/" <> owner <> "/" <> repo
+        appVeyorBadge :: Text = 
+            makeBadge "Windows build status" appVeyorShield appVeyorLink
 
-        licenseShield :: Text =
-          "https://img.shields.io/badge/license-" <> T.replace "-" "--" license <> "-blue.svg"
-        licenseLink :: Text =
-          "https://github.com/" <> owner <> "/" <> repo <> "/blob/master/LICENSE"
+        makeBadge :: Text -> Text -> Text -> Text
+        makeBadge title shield link = "[![" <> title <> "](" <> shield <> ")](" <> link <> ")"
```
ChShersh: If you like, you could use `neat-interpolation` in `makeBadge` function:

```haskell
[text|[![$title]($shield)]($link)|]
```

Choose which one you like more :slightly_smiling_face: I'm okay with both choices :+1:

src/Summoner/Template.hs
```diff
@@ -326,31 +332,27 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
...
+            "https://ci.appveyor.com/project/" <> owner <> "/" <> repo
+        appVeyorBadge :: Text = 
+            makeBadge "Windows build status" appVeyorShield appVeyorLink
 
-        licenseShield :: Text =
-          "https://img.shields.io/badge/license-" <> T.replace "-" "--" license <> "-blue.svg"
-        licenseLink :: Text =
-          "https://github.com/" <> owner <> "/" <> repo <> "/blob/master/LICENSE"
+        makeBadge :: Text -> Text -> Text -> Text
+        makeBadge title shield link = [text|[![$title]($shield)]($link)|]
```
ChShersh: I've just discovered that `neat-interpolation` adds newlines. So I see the following output:

```
# foo

[![Hackage](https://img.shields.io/hackage/v/foo.svg)](https://hackage.haskell.org/package/foo)

[![MPL-2.0 license](https://img.shields.io/badge/license-MPL--2.0-blue.svg)](LICENSE)

[![Stackage Lts](http://stackage.org/package/foo/badge/lts)](http://stackage.org/lts/package/foo)

[![Stackage Nightly](http://stackage.org/package/foo/badge/nightly)](http://stackage.org/nightly/package/foo)
```

Looks like we can't use `neat-interpolation` for `makeBadge` function and we should return to previous version with `<>`.

PR #139, opened 2018-08-28
 [#35] Add config option to add stylish file

### LINKED ISSUE #35: Create '.stylish-haskell.yaml' in project
`stylish-haskell` works that way that it applies rules from `.stylish-haskell.yaml` inside project root if it's there. So it's rather convenient to put some `stylish-haskell` config inside created project if you want to follow some style and want your contributors to use your `stylish-haskell` config.

At first I propose to put [Serokell's `.stylish-haskell.yaml`](https://github.com/serokell/serokell-util/blob/master/.stylish-haskell.yaml) which is quite reasonable and convenient.

vrom911:
Resolves #35

### DIFF
--- a/src/Summoner/Config.hs
+++ b/src/Summoner/Config.hs
@@ -30,7 +30,7 @@ import Data.List (lookup)
 import Data.Monoid (Last (..))
 import Generics.Deriving.Monoid (GMonoid, gmemptydefault)
 import Generics.Deriving.Semigroup (GSemigroup, gsappenddefault)
-import Toml (AnyValue (..), BiToml, Key, Prism (..), dimap, (.=))
+import Toml (AnyValue (..), BiMap (..), BiToml, Key, dimap, (.=))
 
 import Summoner.Decision (Decision (..))
 import Summoner.GhcVer (GhcVer (..), parseGhcVer, showGhcVer)
@@ -155,10 +155,10 @@ configT = Config
     lastT :: (Key -> BiToml a) -> Key -> BiToml (Last a)
     lastT = Toml.wrapper . Toml.maybeT
 
-    _GhcVer :: Prism AnyValue GhcVer
-    _GhcVer = Prism
-        { preview = \(AnyValue t) -> Toml.matchText t >>= parseGhcVer
-        , review = AnyValue . Toml.Text . showGhcVer
+    _GhcVer :: BiMap AnyValue GhcVer
+    _GhcVer = BiMap
+        { forward = \(AnyValue t) -> Toml.matchText t >>= parseGhcVer
+        , backward = Just . AnyValue . Toml.Text . showGhcVer
         }
 
     ghcVerArr :: Key -> BiToml [GhcVer]
--- a/src/Summoner/Config.hs
+++ b/src/Summoner/Config.hs
@@ -30,7 +30,7 @@ import Data.List (lookup)
 import Data.Monoid (Last (..))
 import Generics.Deriving.Monoid (GMonoid, gmemptydefault)
 import Generics.Deriving.Semigroup (GSemigroup, gsappenddefault)
-import Toml (AnyValue (..), BiMap (..), BiToml, Key, dimap, (.=))
+import Toml (AnyValue (..), BiMap (..), BiToml, Bijection (..), Key, dimap, (.=))
 
 import Summoner.Decision (Decision (..))
 import Summoner.GhcVer (GhcVer (..), parseGhcVer, showGhcVer)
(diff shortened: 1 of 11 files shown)


src/Summoner/Template.hs
```diff
@@ -433,7 +434,7 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
         |]
       where
         githubLine :: Text = memptyIfFalse github $ "The change log is available [on GitHub][2]."
-        githubFootNote :: Text = memptyIfFalse github $
```
willbasky: So many spaces I left. What tool show them to you?

willbasky:
Does stylish file created when there is field inside `.toml` only or while interactive mode too?

src/Summoner/Template.hs
vrom911: @willbasky I've set up my emacs to remove this for me automatically.

src/Summoner/Config.hs
```diff
@@ -151,14 +154,15 @@ configT = Config
...
     lastT = Toml.wrapper . Toml.maybeT
 
-    _GhcVer :: Prism AnyValue GhcVer
-    _GhcVer = Prism
-        { preview = \(AnyValue t) -> Toml.matchText t >>= parseGhcVer
-        , review = AnyValue . Toml.Text . showGhcVer
+    _GhcVer :: BiMap AnyValue GhcVer
+    _GhcVer = BiMap
+        { forward = \(AnyValue t) -> Toml.matchText t >>= parseGhcVer
+        , backward = Just . AnyValue . Toml.Text . showGhcVer
```
ChShersh: There should be easier way to create such `BiMap` in `tomland`. It's not implemented currently but looks like we need it.

src/Summoner/Source.hs
```diff
@@ -0,0 +1,55 @@
...
+
+module Summoner.Source
+       ( Source (..)
+       , sourceT
+       , fetchSource
+       ) where
+
+import Relude
+
+import Control.Arrow ((>>>))
```
ChShersh: What do you think if `relude` exports `>>>` and `<<<`? 🤔 
Looks like this operators are very common. And it has unambiguous name.

src/Summoner/Source.hs
```diff
@@ -0,0 +1,55 @@
...
+    -- | File path to the local source file.
+    | File FilePath
+    deriving (Show, Eq)
+
+sourceT :: Key -> BiToml Source
+sourceT nm = Toml.match (Toml._Text   >>> _Url)  (nm <> "url")
+         <|> Toml.match (Toml._String >>> _File) (nm <> "file")
+  where
+    _Url :: BiMap Text Source
+    _Url = Toml.invert $ Toml.prism (source Just (const Nothing)) Url
```
ChShersh: We need to add some utility function to `tomland` which is like `Toml.invert . Toml.prism` but I don't know good name for such function... 😞 Looks like `prism` function is not needed in most cases and often we need `Toml.invert . Toml.prism`.

src/Summoner/Source.hs
```diff
@@ -0,0 +1,55 @@
...
+    _File = Toml.invert $ Toml.prism (source (const Nothing) Just) File
+
+    source :: (Text -> c) -> (FilePath -> c) -> Source -> c
+    source f _ (Url x)  = f x
+    source _ f (File x) = f x
+
+fetchSource :: Source -> IO (Maybe Text)
+fetchSource = \case
+    File path -> catch (Just <$> readFile path) (fileError path)
+    Url url -> catch (fetchUrl url) (urlError url)
```
ChShersh: You can write this like:

```haskell
fetchUrl url `catch` urlError url
```

But I'm not sure whether it will work with `<$>`

summoner.cabal
```diff
@@ -65,7 +66,7 @@ library
                      , process
                      , text
                      , time
-                     , tomland ^>= 0.3.1
+                     , tomland ^>= 0.4.0
                      , relude ^>= 0.1.1
```
ChShersh: I think we can also migrate to `relude ^>= 0.2.0`. This should be an easy migration.

src/Summoner/Config.hs
```diff
@@ -151,14 +154,15 @@ configT = Config
     <*> lastT (Toml.table preludeT) "prelude" .= cPrelude
     <*> textArr         "extensions"  .= cExtensions
     <*> textArr         "warnings"    .= cWarnings
+    <*> lastT sourceT   "stylish"     .= cStylish
```
ChShersh: Okay, I see problem in this parser. Looks like `maybeT` parser that I've implemented doesn't work 😞 It has the following type:

```haskell
maybeT :: forall a . (Key -> BiToml a) -> Key -> BiToml (Maybe a)
```

The problem is in the fact that if either `stylish.url` or `stylish.file` absent, `maybeT` checks whether exception is `KeyNotFound "stylish"` or `TableNotFound "stylish"`. But the actual exception is none of this! So looks like `maybeT` combinator is completely broken.

I propose to remove `maybeT` completely from `tomland` in next major release. Since `BiToml` has an `Alternative` instance, you can use [`optional`](https://hackage.haskell.org/package/base-4.11.1.0/docs/Control-Applicative.html#v:optional) to achieve semantic of `maybeT`. 

Currently we have:

```haskell
    <*> lastT sourceT   "stylish"     .= cStylish

lastT :: (Key -> BiToml a) -> Key -> BiToml (Last a)
lastT = Toml.wrapper . Toml.maybeT
```

The fix is to write like this:

```haskell
    <*> wrapLastT (maybeSourceT "stylish") .= cStylish

wrapLastT :: BiToml (Maybe a) -> BiToml (Last a)
wrapLastT = Toml.dimap getLast Last

maybeSourceT :: Key -> BiToml (Maybe Source)
maybeSourceT key = optional (sourceT key)
```

Issue #155, opened 2018-09-30
Migrate to tomland v0.5.0


ch3rn0v:
Hello. I decided to tackle a couple Hacktoberfest issues here, and I encounter this error during dependencies installation.

`cabal install` produces the following output:

```
src/Toml/Parser.hs:30:33: error:
    Module ‘Text.Megaparsec’ does not export ‘parseErrorPretty'’
   |
30 | import Text.Megaparsec (Parsec, parseErrorPretty', try)
   |                                 ^^^^^^^^^^^^^^^^^

src/Toml/Parser.hs:31:44: error:
    Module ‘Text.Megaparsec.Char’ does not export ‘anyChar’
   |
31 | import Text.Megaparsec.Char (alphaNumChar, anyChar, char, digitChar, eol, hexDigitChar, oneOf,
   |                                            ^^^^^^^

src/Toml/Parser.hs:31:89: error:
    Module ‘Text.Megaparsec.Char’ does not export ‘oneOf’
   |
31 | import Text.Megaparsec.Char (alphaNumChar, anyChar, char, digitChar, eol, hexDigitChar, oneOf,
   |                                                                                         ^^^^^

src/Toml/Parser.hs:32:30: error:
    Module ‘Text.Megaparsec.Char’ does not export ‘satisfy’
   |
32 |                              satisfy, space, space1, string, tab)
   |                              ^^^^^^^
```

So I cloned the `tomland` package and installed it from source. As the version number suggests, I didn't expect backward compability and that expectation was met:

```
src/Summoner/Source.hs:10:21: error:
    Module ‘Toml’ does not export ‘BiToml’
   |
10 | import Toml (BiMap, BiToml, Key)
   |                     ^^^^^^
```

Is it just me or the `summoner` project should be updated according to the new version of `tomland`?

P. S. Yes, I checked out the `v0.4.0` tag. I didn't manage to build this version at all, `cabal install` command results in the same error as in the first output of this message.

PR #157, opened 2018-09-30
[#156] Add build-depends constraint

### LINKED ISSUE #156: Add build-depends bounds

ChShersh:
Resolves #156 

This is a temporary fix. Should be reverted after migration to newer `tomland` version (see #155). Also, we should add more bounds later.


ChShersh:
Even better: I've updated `tomland` via revision, so we shouldn't need `megaparsec` constraint. In this PR I've just added constraints to other dependencies. I think this is useful.

PR #158, opened 2018-10-01
[#154] - Add Link constructor to Source data type

### LINKED ISSUE #154: Add `Link` constructor to `Source` data type
We have `Source` data type that allows to create files in repository by copying file from filesystem or by downloading from URL. It's often desired to have 3rd case: create small file with the link to existing file. It's useful for organizations to have only single `CONTRIBUTING.md` for example, and then link to this file in every repo instead of duplicating its content.

So this can be implemented by adding `Link` constructor to the following data type:

https://github.com/kowainik/summoner/blob/4395c5f93c4a94d489c0824f733e9127bfdb48a0/src/Summoner/Source.hs#L16-L21

Cmdv:
this fixes #154 
but I have failing tests and I'm not sure I'm carrying it out the right way.
`cContributing` is having a bit of a nightmare but I wonder why would `Link` differ from `Url`?

### DIFF
--- a/src/Summoner/Source.hs
+++ b/src/Summoner/Source.hs
@@ -1,15 +1,18 @@
+{-# LANGUAGE QuasiQuotes     #-}
+
 module Summoner.Source
        ( Source (..)
        , sourceT
        , fetchSource
        ) where
 
-import Control.Arrow ((>>>))
-import Control.Exception (catch)
-import System.Process (readProcess)
-import Toml (BiMap, BiToml, Key)
+import           Control.Arrow ((>>>))
+import           Control.Exception (catch)
+import           NeatInterpolation (text)
+import           System.Process (readProcess)
+import           Toml (BiMap, BiToml, Key)
 
-import Summoner.Ansi (errorMessage)
+import           Summoner.Ansi (errorMessage)
 
 import qualified Toml
 
@@ -18,7 +21,7 @@ data Source
     = Url Text
     -- | File path to the local source file.
     | File FilePath
-    -- | Link to source source file.
+    -- | Link to external file.
     | Link Text
     deriving (Show, Eq)
 
@@ -45,7 +48,7 @@ fetchSource :: Source -> IO (Maybe Text)
 fetchSource = \case
     File path -> catch (Just <$> readFileText path) (fileError path)
     Url url -> catch (fetchUrl url) (urlError url)
-    Link link -> catch putLink (urlError link)
+    Link link -> catch (putLink link) (urlError link)
   where
     fileError :: FilePath -> SomeException -> IO (Maybe Text)
     fileError path _ = errorMessage ("Couldn't read file: " <> toText path)
@@ -58,6 +61,5 @@ fetchSource = \case
     fetchUrl :: Text -> IO (Maybe Text)
     fetchUrl url = Just . toText <$> readProcess "curl" [toString url] ""
 
-    putLink :: IO (Maybe Text)
-    putLink = errorMessage "See this document under the [following link](<link goes here>)"
-              >> pure Nothing
+    putLink :: Text -> IO (Maybe Text)
+    putLink link = pure $ Just [text|See full content of the file [here]($link)|]
--- a/src/Summoner/Source.hs
+++ b/src/Summoner/Source.hs
@@ -18,11 +18,14 @@ data Source
     = Url Text
     -- | File path to the local source file.
     | File FilePath
+    -- | Link to source source file.
+    | Link Text
     deriving (Show, Eq)
 
 sourceT :: Key -> BiToml Source
 sourceT nm = Toml.match (Toml._Text   >>> _Url)  (nm <> "url")
          <|> Toml.match (Toml._String >>> _File) (nm <> "file")
+         <|> Toml.match (Toml._Text >>> _Link) (nm <> "link")
   where
     _Url :: BiMap Text Source
     _Url = Toml.invert $ Toml.prism (source Just (const Nothing)) Url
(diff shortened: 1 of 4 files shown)


src/Summoner/Source.hs
```diff
@@ -18,26 +18,34 @@ data Source
...
     source :: (Text -> c) -> (FilePath -> c) -> Source -> c
     source f _ (Url x)  = f x
     source _ f (File x) = f x
+    source f _ (Link x) = f x
 
 fetchSource :: Source -> IO (Maybe Text)
 fetchSource = \case
     File path -> catch (Just <$> readFileText path) (fileError path)
     Url url -> catch (fetchUrl url) (urlError url)
+    Link link -> catch (fetchUrl link) (urlError link)
```
ChShersh: Here you should not fetch the link. But instead put some text like:

```markdown
See this document under the [following link](<link goes here>)
```

Though, it might not work if the file is not formatted with markdown... But I think it's mostly for `.md` files so should work.

src/Summoner/Source.hs
```diff
@@ -18,26 +18,34 @@ data Source
...
     _Url :: BiMap Text Source
     _Url = Toml.invert $ Toml.prism (source Just (const Nothing)) Url
 
     _File :: BiMap FilePath Source
     _File = Toml.invert $ Toml.prism (source (const Nothing) Just) File
 
+    _Link :: BiMap Text Source
+    _Link = Toml.invert $ Toml.prism (source Just (const Nothing)) Link
+
     source :: (Text -> c) -> (FilePath -> c) -> Source -> c
```
ChShersh: I think that tests are failing because of the fact that you handle different constructors with the same function. Try adding one extra parameter to `source` function and write:

```haskell
source _ _ f (Link x) = f x
```

test/Test/TomlSpec.hs
```diff
@@ -59,7 +65,7 @@ genPartialConfig = do
     cExtensions <- genTextArr
     cWarnings   <- genTextArr
     cStylish    <- Last <$> Gen.maybe genSource
-    cContributing <- Last <$> Gen.maybe genSource
+    cContributing <- Last <$> Gen.maybe genSourceContrib
```
ChShersh: I think it's better to patch `genSource` function and use it here as well.

src/Summoner/Source.hs
```diff
@@ -49,3 +57,7 @@ fetchSource = \case
 
     fetchUrl :: Text -> IO (Maybe Text)
     fetchUrl url = Just . toText <$> readProcess "curl" [toString url] ""
+
+    putLink :: IO (Maybe Text)
+    putLink = errorMessage "See this document under the [following link](<link goes here>)"
```
Cmdv: @ChShersh I don't quite think I have this right or if it's what you meant in your previous comment?

src/Summoner/Source.hs
```diff
@@ -18,26 +18,34 @@ data Source
...
 
     _File :: BiMap FilePath Source
-    _File = Toml.invert $ Toml.prism (source (const Nothing) Just) File
+    _File = Toml.invert $ Toml.prism (source (const Nothing) Just (const Nothing)) File
 
-    source :: (Text -> c) -> (FilePath -> c) -> Source -> c
-    source f _ (Url x)  = f x
-    source _ f (File x) = f x
+    _Link :: BiMap Text Source
+    _Link = Toml.invert $ Toml.prism (source (const Nothing) (const Nothing) Just) Link
```
ChShersh: I hope that with new `tomland` release there will be better way to write this code but should work for now!

src/Summoner/Source.hs
```diff
@@ -18,26 +18,34 @@ data Source
     = Url Text
     -- | File path to the local source file.
     | File FilePath
+    -- | Link to source source file.
```
ChShersh: Double `source` here. Probably a typo :slightly_smiling_face: Maybe better to say `Link to external file` or something like that? :thinking:

src/Summoner/Source.hs
ChShersh: @Cmdv my ideas was that for the `Link` constructor there should be no IO actions. So you can just write:

```haskell
putLink link = pure $ Just $ "See full content of the file [here](" <> link <> ")"
```

Or you can even use `text` QuasiQuoter from `neat-interpolation` library that we're using in `summoner`:

```haskell
Link link -> pure $ Just [text|See full content of the file [here](${link})|]
```

test/Test/TomlSpec.hs
```diff
@@ -38,12 +38,6 @@ genSource = do
     s   <- Gen.element [File . toString, Url]
```
ChShersh: I think it's a good idea to add `Link` constructor here

src/Summoner/Source.hs
Cmdv: oops! 🤦‍♂️ agreed on term sounds much better.

src/Summoner/Source.hs
Cmdv: great thanks I see what you mean now, `QuasiQuoter` is nice for sure. 👍

Cmdv: actually one question:
`fetchSource :: Source -> IO (Maybe Text)` needs the IO, or should I change that too? because I've gone with:
``` haskell
putLink :: Text -> IO (Maybe Text)
putLink link = pure $ Just [text|See full content of the file [here]($link)|]
```

ChShersh: @Cmdv It's okay to have `pure $ Just`. All right sides of `case ... of` expression should have the same type. So it's okay to add some `pure`s

src/Summoner/Source.hs
```diff
@@ -1,15 +1,18 @@
...
-import System.Process (readProcess)
-import Toml (BiMap, BiToml, Key)
+import           Control.Arrow ((>>>))
+import           Control.Exception (catch)
+import           NeatInterpolation (text)
+import           System.Process (readProcess)
+import           Toml (BiMap, BiToml, Key)
 
-import Summoner.Ansi (errorMessage)
+import           Summoner.Ansi (errorMessage)
```
vrom911: Oops, I think these `stylish-haskell` changes are not intentional 🙂 
(Btw, we have `.stylish-haskell.yaml` in the repo with the config that we are using).

src/Summoner/Source.hs
```diff
@@ -18,26 +21,34 @@ data Source
...
+    source :: (Text -> c) -> (FilePath -> c) -> (Text -> c) -> Source -> c
+    source f _ _ (Url x)  = f x
+    source _ f _ (File x) = f x
+    source _ _ f (Link x) = f x
 
 fetchSource :: Source -> IO (Maybe Text)
 fetchSource = \case
     File path -> catch (Just <$> readFileText path) (fileError path)
     Url url -> catch (fetchUrl url) (urlError url)
+    Link link -> catch (putLink link) (urlError link)
```
vrom911: I guess `catch` is not important in this particular part as the `putLink` function won't produce any error as it's just `pure . Just`.
So you can use just `putLink link` in here.

Cmdv:
@ChShersh cool will do those now, have to say thanks to you guys for taking the time to do all this it's really appreciated. As I look through all your repos and the help you are giving people is superb 👍

vrom911:
@Cmdv Thank you very much for these kind words ❤️ 

About the version: we have already released `1.1.0.1`, so the new changes should go to the `1.2.0`, can you update it in the `.cabal` file as well?

PR #162, opened 2018-10-01
Make GHC 8.6.1 the default one

### LINKED ISSUE #151: Make GHC 8.6.3 the default one for the generated project
Also, we should introduce some logic to take into consideration the fact that Stackage is not that fast.

v0d1ch:
This pr bumps the stack resolver to nightly-2018-10-01 which uses ghc 8.6.1

### DIFF
--- a/stack-8.6.1.yaml
+++ b/stack-8.6.1.yaml
@@ -2,5 +2,4 @@ resolver: nightly-2018-10-01
 
 extra-deps:
   - base-noprelude-4.12.0.0
-  - relude-0.3.0
   - tomland-0.4.0
--- a/stack-8.6.1.yaml
+++ b/stack-8.6.1.yaml
@@ -1,0 +1,6 @@
+resolver: nightly-2018-10-01
+
+extra-deps:
+  - base-noprelude-4.12.0.0
+  - relude-0.3.0
+  - tomland-0.4.0
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,6 +1,6 @@
-resolver: nightly-2018-10-01
+resolver: lts-12.9
 
 extra-deps:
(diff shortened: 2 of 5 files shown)


stack.yaml
```diff
@@ -1,6 +1,6 @@
-resolver: lts-12.9
+resolver: nightly-2018-10-01
```
vrom911: I guess you shouldn't change the `stack.yaml` file as we still want to support `8.4.3` instead you can put this into `stack-8.6.1.yaml` file 🙂

summoner.cabal
```diff
@@ -47,7 +47,7 @@ library
   other-modules:       Paths_summoner
   autogen-modules:     Paths_summoner
 
-  build-depends:       base-noprelude >= 4.10 && < 5
+  build-depends:       base-noprelude >= 4.12 && < 5
```
vrom911: We are not going to give up the support of the `8.2.2` and `8.4.3` yet so no need to upgrade the lower bound. The cost of supporting those versions is very low (we already have everything working in there) but the usability on different versions is very valuable!

v0d1ch:
@vrom911 I added the changes , It should be fine now.

stack-8.6.1.yaml
```diff
@@ -0,0 +1,6 @@
+resolver: nightly-2018-10-01
+
+extra-deps:
+  - base-noprelude-4.12.0.0
+  - relude-0.3.0
```
ChShersh: Actually, `relude-0.3.0` is already in this nightly snapshot, so you can safely remove it from `extra-deps`

v0d1ch: Done!

PR #163, opened 2018-10-02
[153] Add -Werror to build option in generated .travis.yml

Simeon979:


src/Summoner/Template.hs
```diff
@@ -536,15 +536,15 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
...
+              cabal new-build --enable-tests --enable-benchmarks --ghc-options=-Werror
             else
               mkdir -p ~/.local/bin
               export PATH="$$HOME/.local/bin:$$PATH"
               travis_retry curl -L 'https://www.stackage.org/stack/linux-x86_64' | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
               stack --version
               stack setup --no-terminal --install-cabal 2.0.1.0
               stack ghc -- --version
-              stack build --only-dependencies --no-terminal
+              stack build --only-dependencies --no-terminal --ghc-options=-Werror
```
vrom911: this is `--only-dependencies` so no need to add the option here. You should add it here instead:

https://github.com/kowainik/summoner/blob/6d9cc81078d43fc72c959bfe7659c16952710617/src/Summoner/Template.hs#L552

src/Summoner/Template.hs
```diff
@@ -579,7 +579,7 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
           - stack --version
           - stack setup --no-terminal --install-cabal 2.0.1.0
           - stack ghc -- --version
-          - stack build --only-dependencies --no-terminal
+          - stack build --only-dependencies --no-terminal --ghc-options=-Werror
```
vrom911: The same here

src/Summoner/Template.hs
```diff
@@ -579,7 +579,7 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
...
         script:
           - stack build --test --bench --no-run-benchmarks --no-terminal
```
vrom911: The option should be added to this line

src/Summoner/Template.hs
```diff
@@ -536,15 +536,15 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
               export PATH="/opt/ghc/$$GHCVER/bin:/opt/cabal/$$CABALVER/bin:$$PATH"
               echo $$PATH
               cabal new-update
-              cabal new-build --enable-tests --enable-benchmarks
+              cabal new-build --enable-tests --enable-benchmarks --ghc-options=-Werror
```
vrom911: Unfortunately, this doesn't work for `cabal` 😞  See relevant issue: https://github.com/haskell-CI/haskell-ci/issues/95

You can either use the trick mentioned in the issue or just leave `-Werror` only for `stack` and create another issue in `summoner`.

ChShersh: And here is relevant `cabal` issue:

* https://github.com/haskell/cabal/issues/3883

PR #164, opened 2018-10-02
[153] Add -Werror to build option in generated .travis.yml

Simeon979:
an update to [the previous pull request](https://github.com/kowainik/summoner/pull/163)  I closed the previous one because I somehow made a second commit

### DIFF
--- a/src/Summoner/Template.hs
+++ b/src/Summoner/Template.hs
@@ -535,6 +535,8 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
             if [ -z "$$STACK_YAML" ]; then
               export PATH="/opt/ghc/$$GHCVER/bin:/opt/cabal/$$CABALVER/bin:$$PATH"
               echo $$PATH
+              echo "package $PKGNAME" >> cabal.project
+              echo "  ghc options: -Werror" >> cabal.project
               cabal new-update
               cabal new-build --enable-tests --enable-benchmarks
             else
@@ -551,7 +553,7 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
             if [ -z "$$STACK_YAML" ]; then
                ${cabalTest}
             else
-              stack build --test --bench --no-run-benchmarks --no-terminal
+              stack build --test --bench --no-run-benchmarks --no-terminal --ghc-options=-Werror
             fi
         $endLine
         |]
@@ -562,6 +564,8 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
         install:
           - export PATH="/opt/ghc/$$GHCVER/bin:/opt/cabal/$$CABALVER/bin:$$PATH"
           - echo $$PATH
+          - echo "package $PKGNAME" >> cabal.project
+          - echo "  ghc-options: -Werror" >> cabal.project
           - cabal new-update
           - cabal new-build --enable-tests --enable-benchmarks
         script:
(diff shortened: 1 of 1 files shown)


Simeon979:
I guess I have to use double "$" inside quasi quotation?

Simeon979:
Also, will `PKGNAME `need to be present in `env` ?

src/Summoner/Template.hs
```diff
@@ -535,6 +535,8 @@ createProjectTemplate ProjectData{..} = Dir (toString repo) $
             if [ -z "$$STACK_YAML" ]; then
               export PATH="/opt/ghc/$$GHCVER/bin:/opt/cabal/$$CABALVER/bin:$$PATH"
               echo $$PATH
+              echo "package $PKGNAME" >> cabal.project
```
vrom911: Yes, you're right. You need to escape `$` as it's reserved for vars in `neat-interpolation` – so you should use double `$`.

vrom911: Also, I am now thinking that if a user already has `cabal.project` in the repo to make it work together we need to add an empty line before "package $$PKGNAME"

ChShersh: Also, instead of `$PKGNAME` you should insert `$repo` because we don't set `PKGNAME` environment variable in `.travis.yml`

Simeon979:
I've made the changes

Simeon979:
done @ChShersh

vrom911:
@Simeon979 @ChShersh I created the project to check that this working as expected. But unfortunately, it's not 😞 Everything is perfect with `stack`, but `cabal` can't handle this. It needs `cabal.project` file as the addition to the `.cabal` file. Don't know what to do :( We have a few options:

* Don't support this for `.cabal` and let the users to decide when to create the `cabal.project` file when their package is not building with `cabal`
* Wait for `ghc-options:-Werror` in `cabal` to work only on the package itself, not on all dependencies (I don't know when it will be fixed)
* Create `cabal.project` file by the `summoner`. But in this case, I don't know what should be in there (`test: True`, `benchmarks: True`, any other lines to make `cabal` build/test/bench the project). Also, I'm not sure I know the good place to read about when exactly `cabal` stops working, and when we need this `cabal.project` file to be added.

ChShersh:
@Simeon979 @vrom911 I propose the following resolution for the current problem:

* Add `-Werror` only to `stack` on CI in this PR and then merge this PR to `master`
* Create issue in `summoner` repository about supporting `-Werror` for the `cabal` on Travis CI (so we can resolve it later when we figure out the best way to do this)

Supporting `-Werror` for the `cabal-install` requires to create `cabal.project` file by `summoner` because in order to support `-Werror` only on CI there exist single workaround that involves that Haskell project already has proper `cabal.project` file. I don't want to take hasty decisions about generating `cabal.project` for the `cabal-install` just because we want to support `-Werror` on the CI. I think, that decision about generating `cabal.project` should be discussed under corresponding issue where we can decide what to put in this file and how useful it can be. 

Also, introducing `cabal.project`-related logic to the `.travis.yml` file means that `.travis.yml` files generated by `summoner` are not portable: you can't just copy-paste this file from your project to another one and change project name and expect that things will just work. You also need to ensure that you have `cabal.project` file on that package. Portability is a nice property to have. `stack` requires to have `stack.yaml` files because it doesn't work without such files. But `cabal-install` currently can work without `cabal.project`. It's completely okay to generate `cabal.project` only on CI, but if this involves having this file locally, then things become more difficult. And that's why I think that decision about `cabal.project` generation should be thoroughly discussed.

Simeon979:
Okay, so option one? And maybe add a note in the README?

ChShersh:
@Simeon979 Yes, that'll do :+1:

Simeon979:
Check it now @ChShersh  @vrom911 Also I'm not sure if I've gotten the changelog under the right version. The latest version hadn't been added when I added it.

ChShersh:
@Simeon979 You're welcome! Thank you for your contribution!

PR #165, opened 2018-10-04
Add HLint check to Travis CI

qoelet:
Added `.hlint.yaml` that ignores any HLint errors up to this point

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,10 +88,12 @@ install:
 script:
   - |
     if [ -z "$STACK_YAML" ]; then
-       cabal new-test
+      cabal new-test
     else
       stack build --test --no-terminal
     fi
+  - curl https://raw.githubusercontent.com/kowainik/relude/bc23b0fe2c3699f5c202ed04222766dc0d789fd1/.hlint.yaml -o .hlint-relude.yaml
+  - curl -sSL https://raw.github.com/ndmitchell/neil/master/misc/travis.sh | sh -s -- hlint -h .hlint-relude.yaml .
 
 notifications:
   email: false


.travis.yml
```diff
@@ -88,10 +88,11 @@ install:
 script:
   - |
     if [ -z "$STACK_YAML" ]; then
-       cabal new-test
+      cabal new-test
     else
       stack build --test --no-terminal
     fi
+  - curl -sSL https://raw.github.com/ndmitchell/hlint/master/misc/travis.sh | sh -s .
```
ChShersh: @qoelet Thanks for your contribution! Having `hlint` check on CI is really good. Though, we can improve the configuration even better. `summoner` uses `relude` package as an alternative prelude, which means that we can use HLint rules from `relude`:

* https://github.com/kowainik/relude/blob/master/.hlint.yaml

I would appreciate that in Travis CI we download `.hlint.yaml` file for the version of `relude` we're using, rename this file to `.hlint-relude.yaml` (just in case we will have default `.hlint.yaml` in future) and run `hlint` with `.hlint-relude.yaml` hints.

qoelet: @ChShersh updated the PR, let me know if that's OK.

.travis.yml
```diff
@@ -88,10 +88,12 @@ install:
 script:
   - |
     if [ -z "$STACK_YAML" ]; then
-       cabal new-test
+      cabal new-test
     else
       stack build --test --no-terminal
     fi
+  - curl -o .hlint-relude.yaml https://raw.githubusercontent.com/kowainik/relude/master/.hlint.yaml
+  - curl -sSL https://raw.github.com/ndmitchell/neil/master/misc/travis.sh | sh -s -- hlint -h .hlint-relude.yaml .
```
ChShersh: This is really suprising that we don't have any hints in `summoner` and our code is perfect :thinking: :thinking: :thinking: 
But I've checked it locally, looks like we really don't have any hints

.travis.yml
```diff
@@ -88,10 +88,12 @@ install:
 script:
   - |
     if [ -z "$STACK_YAML" ]; then
-       cabal new-test
+      cabal new-test
     else
       stack build --test --no-terminal
     fi
+  - curl -o .hlint-relude.yaml https://raw.githubusercontent.com/kowainik/relude/master/.hlint.yaml
```
ChShersh: It's not enough to fetch from `master` branch because `master` can be ahead of time. You should check the version we use in `summoner` and then use commit from GitHub under corresponding tag. We're using `relude-0.3.0`, it's the latest version currently. So you should take `.hlint.yaml` from the commit tagged by `v0.3.0` version:

* https://github.com/kowainik/relude/releases/tag/v0.3.0

qoelet: Right, that's a good point. I have to dig into my non-existent bash-fu for something :thinking:

.travis.yml
```diff
@@ -88,10 +88,13 @@ install:
 script:
   - |
     if [ -z "$STACK_YAML" ]; then
-       cabal new-test
+      cabal new-test
     else
       stack build --test --no-terminal
     fi
+  - grep 'relude ^' summoner.cabal | awk '{print "https://github.com/kowainik/relude/archive/v"$4".tar.gz"}' | xargs curl -sSL | tar xvz -C '/tmp'
+  - cp $(grep 'relude ^' summoner.cabal | awk '{print "/tmp/relude-"$4"/.hlint.yaml"}') .hlint-relude.yaml
```
ChShersh: @qoelet Could you please explain, what is going on here? I'm not a bash expert, so explanation of commands is highly appreciated! :slightly_smiling_face:

.travis.yml
qoelet: That's the outcome for me as well - I copied the `.hlint.yaml` from `relude` and it reports "No hints".

.travis.yml
qoelet: @ChShersh I'm not sure if this would be the best solution going forward yet. I'm grepping the cabal file for the line where `relude ^= x.2x.x`, then grabbing the release version to construct the URL to the tarball, and fetch/untar it to `/tmp`. I then copy the HLint rule file over to the project dir. I'm going to see if/search for a more elegant solution if there is one.

ChShersh: @qoelet Keep us updated with the results of your search :+1:

qoelet: @ChShersh there's something I'm reading up on called a sparse checkout in Git; it might be a lot clearer if we use that to show we are fetching the HLint rules off a specific release from relude instead of this bash magic.

ChShersh:
@qoelet At this point it just seems easier to pin specific commit from the `relude` GitHub repo and update each time we bump up `relude` version... Even if you specify version like `^>= 0.3.0`, `cabal` still can use version like `0.3.1`, so taking the version from `.cabal` file is not a reliable way.

qoelet:
@ChShersh Good point, thanks for pointing that out. I'll amend the PR.

vrom911:
@qoelet, any updates on this? 🙂

qoelet:
@vrom911 @ChShersh updated PR. I've moved the steps for grabbing the HLint rules from relude into a separate bash script to make it a bit more clear what is better done.

qoelet:
@ChShersh agreed! :+1:

PR #166, opened 2018-10-05
WIP: Add simple Nix support

qoelet:
Trying a simple approach to adding bare minimum Nix support to a created project without invoking something external like `cabal2nix`, which could be feasible. Instead I'm thinking if we reuse the information already present in the cabal file generator, then we have enough for a working Nix derivation:

```shell
$ summon new foo
  Default config /home/kenny/.summoner.toml file is missing.
Add cabal? [y]/n
  ->   
  Cabal will be added to the project
Add stack? [y]/n
  ->   
  Stack will be added to the project
Add Nix support? [y]/n
  ->   
  Nix support will be added to the project
Repository owner:  [kowainik]
...

$ cd foo

$ nix-build
these derivations will be built:
  /nix/store/iig7f13cxc85pprcqikah1vh3x8q31qz-foo-0.0.0.drv
building '/nix/store/iig7f13cxc85pprcqikah1vh3x8q31qz-foo-0.0.0.drv'...
setupCompilerEnvironmentPhase
Build with /nix/store/sx4pjh0nh0acqszld5rrxwbjnzl3djrg-ghc-8.4.3.
ignoring (possibly broken) abi-depends field for packages
unpacking sources
...

$ ./result/bin/foo
someFunc

```

Thoughts?

### DIFF
--- a/src/Summoner/CLI.hs
+++ b/src/Summoner/CLI.hs
@@ -186,6 +186,7 @@ newP = do
     ignoreFile  <- ignoreFileP
     cabal   <- cabalP
     stack   <- stackP
+    nix     <- nixP
     with    <- optional withP
     without <- optional withoutP
     file    <- optional fileP
@@ -197,6 +198,7 @@ newP = do
             { cPrelude = Last $ Prelude <$> preludePack <*> preludeMod
             , cCabal = cabal
             , cStack = stack
+            , cNix = nix
             }
 
 targetsP ::  Decision -> Parser PartialConfig
@@ -311,6 +313,11 @@ stackP :: Parser Decision
 stackP = flag Idk Yes
        $ long "stack"
       <> help "Stack support for the project"
+
+nixP :: Parser Decision
+nixP = flag Idk Yes
+     $ long "nix"
+     <> help "Nix support for the project"
 
 ----------------------------------------------------------------------------
 -- Beauty util
--- a/src/Summoner/Config.hs
+++ b/src/Summoner/Config.hs
@@ -48,6 +48,7 @@ data ConfigP (p :: Phase) = Config
     , cGhcVer       :: p :- [GhcVer]
     , cCabal        :: Decision
     , cStack        :: Decision
+    , cNix          :: Decision
     , cGitHub       :: Decision
     , cTravis       :: Decision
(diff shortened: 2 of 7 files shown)


willbasky:
I am following [Gabriel](https://github.com/Gabriel439/haskell-nix/tree/master/project1) tutorial just now. And there is added some dependencies to `.nix` file. `.nix` file is main configuration for project. What is generated after 'Nix support will be added to the project`?

qoelet:
@willbasky in this case, a derivation `<repo>.nix` and an accompanying `default.nix`

willbasky:
If we need to add some libraries out of resolver, we just add it to `extradep` in `stack.yaml`. But for `Nix` we should generate `extralib.nix` with `cabal2nix` and then add 2 strings to `release.nix` file to point out `extralib.nix`. And if it won`t compile, it asks for some other extra dependencies and so on.

ChShersh:
@qoelet @willbasky @vrom911 I think that supporting Nix might be more difficult than we think :thinking: Maybe it's better to refactor our code first and think about better modularization. At this moment it's pretty hard to add new significant improvements to the `summoner` without refactoring existing code base into smaller pieces. Also. regarding `nix`: we should somehow specify all versions of libraries in `.nix` file and there're different tools for different build systems: `cabal2nix` and `stack2nix`. So this looks like a rabbit hole at the moment :rabbit2:

qoelet:
@ChShersh Agreed

mgttlinger:
Would it make sense to at least add basic nix file generation using `callCabal2nix`? I mean sure it is a pain point that you lose control over the exact dependency versions used by giving that choice to nix but for my personal projects I'm fine with that. So having this basic support would at least help me while being really low maintenance effort given that we only need to generate two files where only two words need to be changed.

mgttlinger:
So my proposal with using `callCabal2nix` has the problem that e.g. for the summoner project the specified version requirements are not met and would need extensive tweaking by the user. The closest thing to what we want is probably what [stack2nix](https://github.com/input-output-hk/stack2nix) does which generates a nix file specifying the exact dependency versions to use. However currently this also doesn't work for the summoner project [see this issue I opened](https://github.com/input-output-hk/stack2nix/issues/136).

PR #175, opened 2018-10-19
Add .hlint.yaml.

StevenXL:
This is the HLint configuration file generated by running:

`hlint . --default > .hlint.yaml`

Notice that we are ignoring one hint:

`ignore: {name: "Parse error"}`

However, the source code does not actually contain a Parse error. Both
`ghcid` and `stack ghci` can load the source code fine.

I'd like to get rid of this ignore, but it is not as dangerous as it
might seem since, if the source code does result in a Parse error, other
checks would fail loudly.

Resolves #{PUT_ISSUE_NUMBER_HERE}

## Checklist:

- [ ] I've updated the [CHANGELOG](https://github.com/kowainik/summoner/blob/master/CHANGELOG.md) with the short description of my latest changes.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,0 +1,64 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
+# This file contains a template configuration file, which is typically
+# placed as .hlint.yaml in the root of your project
+
+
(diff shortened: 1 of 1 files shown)


StevenXL:
@ChShersh Thank you for the clarification. I'm going to close this down and try again!

PR #183, opened 2018-10-26
`settingsTestedVersions` is of type `NonEmpty GhcVer`.

### LINKED ISSUE #179: Make list of GHC versions `NonEmpty`
List of GHC versions actually never can't be empty. So let's make it `NonEmpty`:

* https://github.com/kowainik/summoner/blob/0a533d01bf1800a4fbe1323f40efdb66a8bbae0b/src/Summoner/Settings.hs#L33

StevenXL:
Resolves #179 
I would love feedback on whether this is idiomatic use of the `NonEmpty` data type. While the code builds and the tests pass, there is a lot of conversion from `NonEmpty` to a list via `toList`. I am wondering if I am missing the forest for the trees here. 

`memptyIfFalse` requires a list since `NonEmpty` does not have a `Monoid` instance; `T.concat` requires a list as well; the only area where I made the function a bit more general is in `intercalateMap`, where we take any instance of `Traversable`, instead of hard-coding the list type. We still need to convert it to a list because we use `T.concat`.

## Checklist:

- [x] I've updated the [CHANGELOG](https://github.com/kowainik/summoner/blob/master/CHANGELOG.md) with the short description of my latest changes.

### DIFF
--- a/src/Prelude.hs
+++ b/src/Prelude.hs
@@ -6,6 +6,7 @@ module Prelude
 
        , endLine
        , memptyIfFalse
+       , toListMap
        ) where
 
 import Relude
@@ -16,3 +17,6 @@ endLine = "\n"
 
 memptyIfFalse :: Monoid m => Bool -> m -> m
 memptyIfFalse p val = if p then val else mempty
+
+toListMap :: Foldable f => (a -> b) -> f a -> [b]
+toListMap f = map f . toList
--- a/src/Summoner/Project.hs
+++ b/src/Summoner/Project.hs
@@ -23,6 +23,7 @@ import Summoner.Source (fetchSource)
 import Summoner.Template (createProjectTemplate)
 import Summoner.Text (intercalateMap, packageToModule)
 import Summoner.Tree (showTree, traverseTree)
+import qualified Data.List.NonEmpty as NE
 
 -- | Generate the project.
 generateProject :: Text -> Config -> IO ()
@@ -71,7 +72,7 @@ generateProject projectName Config{..} = do
     let settingsWarnings = cWarnings
 
     putTextLn $ "The project will be created with the latest resolver for default GHC-" <> showGhcVer defaultGHC
-    settingsTestedVersions <- sortNub . (defaultGHC :) <$> case cGhcVer of
+    settingsTestedVersions <- NE.sort . NE.nub . (defaultGHC :|) <$> case cGhcVer of
         [] -> do
             putTextLn "Additionally you can specify versions of GHC to test with (space-separated): "
             infoMessage $ "Supported by 'summoner' GHCs: " <> intercalateMap " " showGhcVer universe
--- a/src/Summoner/Template/Stack.hs
+++ b/src/Summoner/Template/Stack.hs
@@ -12,7 +12,7 @@ import Summoner.Tree (TreeFs (..))
 
 
 stackFiles :: Settings -> [TreeFs]
-stackFiles Settings{..} = map createStackYaml settingsTestedVersions
+stackFiles Settings{..} = memptyIfFalse settingsStack $ toListMap createStackYaml settingsTestedVersions
  where
     -- create @stack.yaml@ file with LTS corresponding to specified ghc version
     createStackYaml :: GhcVer -> TreeFs
--- a/src/Summoner/Text.hs
+++ b/src/Summoner/Text.hs
@@ -15,8 +15,8 @@ packageToModule = tconcatMap headToUpper . T.splitOn "-"
 
 -- | Converts every element of list into 'Text' and then joins every element
 -- into single 'Text' like 'T.intercalate'.
-intercalateMap :: Text -> (a -> Text) -> [a] -> Text
-intercalateMap between showT = T.intercalate between . map showT
+intercalateMap :: Foldable t => Text -> (a -> Text) -> t a -> Text
+intercalateMap between showT = T.intercalate between . toListMap showT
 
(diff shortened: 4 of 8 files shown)


StevenXL:
@ChShersh Thank you for taking the time to review the code. This is precisely the kind of feedback I was looking / hoping for in order to level up my Haskell skills. I believe I addressed all your concerns, except for 1, which is to create a `delete` function for `NonEmpty a`. My thoughts are that the reason why there is no such function is that what if `NonEmpty` has only one item, and it is the item that we want to delete? Then we'd no longer be able to guarantee a non-empty list. The function would then have to return a `Maybe (NonEmpty a)` to indicate failure to delete?

If you still think it makes sense to create a `delete` function, I'd be happy to take another look. And of course, any other comments are welcomed.

src/Prelude.hs
```diff
@@ -16,3 +17,6 @@ endLine = "\n"
 
 memptyIfFalse :: Monoid m => Bool -> m -> m
 memptyIfFalse p val = if p then val else mempty
+
+mapAndList :: (Foldable t, Functor t) => (a -> b) -> t a -> [b]
+mapAndList f = toList . fmap f
```
ChShersh: I feel like writing this function in the following way is better:

```haskell
toListMap :: Foldable f => (a -> b) -> f a -> [b]
toListMap f = map f . toList
```

In this case you're able to use `map` function specialised to the list data type which potentially can be more performant (doesn't matter in our case, but still). Also, it will make constraints less strict.

src/Summoner/Template/Stack.hs
```diff
@@ -12,7 +12,7 @@ import Summoner.Tree (TreeFs (..))
 
 
 stackFiles :: Settings -> [TreeFs]
-stackFiles Settings{..} = map createStackYaml settingsTestedVersions
+stackFiles Settings{..} = mapAndList createStackYaml settingsTestedVersions
```
ChShersh: The implementation is now wrong here. You still need to add `memptyIfFalse settingsStack` because `stackFiles` function always works with `NonEmpty` list so the output list will be always non-empty as well. But it should be empty if users disabled `stack` build tool.

src/Summoner/Text.hs
```diff
@@ -14,8 +14,8 @@ packageToModule = T.concat . map headToUpper . T.splitOn "-"
 
 -- | Converts every element of list into 'Text' and then joins every element
 -- into single 'Text' like 'T.intercalate'.
-intercalateMap :: Text -> (a -> Text) -> [a] -> Text
-intercalateMap between showT = T.intercalate between . map showT
+intercalateMap :: (Functor t, Foldable t) => Text -> (a -> Text) -> t a -> Text
+intercalateMap between showT = T.intercalate between . mapAndList showT
```
ChShersh: Also, relaxing constraints for `mapAndList` function will allow to relax constraints for the `intercalateMap` function which will this function from using `[a]` to using some `Foldable f => f a` and this is an idiomatic generalisation for such kind of things.

src/Summoner/Project.hs
```diff
@@ -71,7 +73,7 @@ generateProject projectName Config{..} = do
     let settingsWarnings = cWarnings
 
     putTextLn $ "The project will be created with the latest resolver for default GHC-" <> showGhcVer defaultGHC
-    settingsTestedVersions <- sortNub . (defaultGHC :) <$> case cGhcVer of
+    settingsTestedVersions <- NE.sortWith id . nub . (defaultGHC :|) <$> case cGhcVer of
```
ChShersh: You can use `sort` function for `NonEmpty`. And, also, let's make call to `nub` qualified, so it will be something like:

```haskell
NE.sort . NE.nub
```

StevenXL:
@ChShersh Thanks again for the feedback. I don't know why I didn't find `NE.sort`.

vrom911:
This PR looks very nice, thanks @StevenXL !

I was thinking about this issue more and I have another idea 🤔 (sorry for not saying it during discussion times, but I only see it now). 
I noticed another approach which can probably reduce some complexity in our code with ghc versions. What if instead of having `NonEmpty` we will have separately the default version of the GHC and the list (which apparently can be empty) of the user-chosen GHCs?
In this case we don't need to dance with removing `defaultGhc` from the list in some places, and with special pattern matching on `GHC-8.4.3` (which is the default at the moment).

What do you think?

ChShersh:
@vrom911 Wow, that's a good point! I didn't even think about this. If you think more, it actually makes much more sense: currently it looks kinda weird to add `defaultGhc` to the list and them remove it from this list. I feel that such type of refactorings allow you to understand Haskell types better: sure, it's good to have more type-safety guarantees, but in our case we actually don't get benefits from converting list to `NonEmpty` 😞 Because we work with `NonEmpty` just as with ordinary list, and having `NonEmpty` doesn't improve code (like allows us to use `head` for `NonEmpty` instead of `Unsafe.head` for list). 

On the other hand, it wasn't clear to me that `NonEmpty` won't give us any benefits. Sometimes you see things only after they are implemented. So such work is very valuable 👍 Not everything needs to be merged to `master` to be extremely valuable.

I can propose the following resolution to this problem:
1. Close the issue with moving `defaultGhc` to the `Settings` (or something like that) or rename the issue and write new description.
2. Close this PR or remake it under new context for the issue.

StevenXL:
@vrom911 That's a fantastic idea, and I'm all for it. @ChShersh Could we proceed by closing this PR (and the associated issue), and opening up a new issue that is assigned to me?

Issue #187, opened 2018-10-27
[RFC] TUI interface for summoner


vrom911:
# Proposal
Looking at the amount of the questions in the interactive mode of `summoner` I feel a bit unsatisfied. To get rid of that I propose to make `summoner` a terminal app with TUI. So you will see all settings on one screen where you'll be able to select the checkboxes  or to edit names/other text options like in CLI that you'd like to see in the generated project.

Note, that it should work absolutely identical to what we have now with the only difference – `summoner` appearance (something like this very rough sketch):

```
┌─────────────  Summoner  ───────────────┐
│                                        │
│ Project name: [                ]       │
│                                        │
│                    ...                 │
│                                        │
│┌───────────  Build Tool  ─────────────┐│
││[x] Cabal                             ││
││[x] Stack                             ││
││[ ] Nix                               ││
│└──────────────────────────────────────┘│
│                ...                     │
└────────────────────────────────────────┘
```

We can use [`brick`]() librarary for this.

## Pros

* More flexible interface
* Ability to see all settings you can change immediatly
* On-fly check of invalid fields
* Easy edit of last steps (which is not possible in CLI, see #177)
* Possibility to easy add features (just use your imagination)

## Cons

Unfortunately, this feature has some drawbacks. 

* __Important:__ Windows support
  _Solution:_ It's not good to just stop supporting this OS, so the solution is to have `summoner-cli` and `summoner-tui` as internal libraries, so anyone can at least install the tool. Also, there will be two executables: `summon` and `summon-tui` (or any other names), the first one will depend on the main `summoner-cli` library (which is the one we have now) and the second one will run TUI interface for summoner. 
* Few internal libraries to maintain
* Complicated installation process (Use some flags for installing the certain library (CLI or TUI))
* `brick` is quite heavy dependency

---

Basically, we are not going to give up everything we have now, but will add feature with more robust interface for users who is comfortable with it.

Any comments, thought, other drawbacks are appreciated.

ChShersh:
I really like the idea :heart_eyes: I see that there're a lot of crazy possibilities that we can do with TUI:

1. Display project directory content on-the-fly as you change settings.
2. Add possibility to view or even change content of the generated files (for example, if you want to edit `.cabal` file a little bit or `.travis.yml`).

And much more. I expect that it will require a lot of time to implement and support all features, but I really like the idea :100: :top:

ChShersh:
[...]

PR #194, opened 2018-11-02
[#193] Do not store defaultghc in tested versions

### LINKED ISSUE #193: Not store `defaultGHC` in the list of tested-versions
See comments started from here:

* https://github.com/kowainik/summoner/pull/183#issuecomment-434916360

StevenXL:
Resolves #193 

## Checklist:

- [x] I've updated the [CHANGELOG](https://github.com/kowainik/summoner/blob/master/CHANGELOG.md) with the short description of my latest changes.
- [x] I changed code related to the generated project.

### DIFF
--- a/src/Summoner/Template/Stack.hs
+++ b/src/Summoner/Template/Stack.hs
@@ -9,11 +9,13 @@ import NeatInterpolation (text)
 import Summoner.GhcVer (GhcVer (..), baseVer, latestLts, showGhcVer)
 import Summoner.Settings (Settings (..))
 import Summoner.Tree (TreeFs (..))
+import Summoner.Default (defaultGHC)
 
 
 stackFiles :: Settings -> [TreeFs]
-stackFiles Settings{..} = map createStackYaml settingsTestedVersions
+stackFiles Settings{..} = map createStackYaml ghcVersions
  where
+    ghcVersions = sortNub (defaultGHC : settingsTestedVersions)
     -- create @stack.yaml@ file with LTS corresponding to specified ghc version
     createStackYaml :: GhcVer -> TreeFs
     createStackYaml ghcV = File (toString $ "stack" <> ver <> ".yaml")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 1.2.0
 =======
 
+* [#193](https://github.com/kowainik/summoner/issues/193):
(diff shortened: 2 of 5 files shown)


src/Summoner/Template/Stack.hs
```diff
@@ -9,11 +9,13 @@ import NeatInterpolation (text)
 import Summoner.GhcVer (GhcVer (..), baseVer, latestLts, showGhcVer)
 import Summoner.Settings (Settings (..))
 import Summoner.Tree (TreeFs (..))
+import Summoner.Default (defaultGHC)
 
 
 stackFiles :: Settings -> [TreeFs]
-stackFiles Settings{..} = map createStackYaml settingsTestedVersions
+stackFiles Settings{..} = map createStackYaml ghcVersions
```
ChShersh: `memptyIfFalse settingsStack` still should be used here because `ghcVersions` is not an empty list and we don't wont to generate `stack.yaml` files if user didn't ask for them 🙂

src/Summoner/Template/Stack.hs
```diff
@@ -9,11 +9,13 @@ import NeatInterpolation (text)
...
  where
+    ghcVersions = sortNub (defaultGHC : settingsTestedVersions)
```
ChShersh: I think that one of the point of this refactoring is to remove explicit pattern-matching on `Ghc843` and compare GHC version with the `defaultGHC` in `ver` variable below.

ChShersh:
@StevenXL After discussion with @vrom911 we decide to go one step further. Let's ask about default GHC at runtime via separate question and then ask for additional GHC versions. This will become useful with #198 because some people might want to use GHC-8.4.3 as a default but some might want to use GHC-8.4.4.

ChShersh:
@StevenXL The structure of the project changed due to #187. Summoner now is a multipackage project. That's why you can see conflicts in all your files. Since the issue has changed (again; yeah, sorry for that, the issue with the `defaultGHC` turned out to be much more difficult from the UX experience), maybe it's easier to close this PR and start again. 

I want to say sorry again, I haven't foreseen that this issue might give us a lot of troubles... If you have any questions regarding the issue, please, don't hesitate to ask :slightly_smiling_face:

StevenXL:
@ChShersh It's not a problem at all. Sorry that I have not been responsive - I was on vacation and I am back now but catching up with regular work. I think it is best to close this PR. Thank you !

PR #198, opened 2018-11-06
[#197] Update lts and other settings for GHC-8.4.4

### LINKED ISSUE #197: Update LTS resolver to 12.18 and GHC to 8.4.4 for stack
For both: generated project and `summoner` itself

* https://www.stackage.org/lts-12.17

cronokirby:
Resolves #197 

Changes [travis.yml](https://github.com/kowainik/summoner/blob/master/.travis.yml)
and [stack.yaml](https://github.com/kowainik/summoner/blob/master/stack.yaml)
to use GHC 8.4.4 as well.

Also adds support for using GHC 8.4.4 in project generation, and makes it the default.

## Checklist:

- [X] I've updated the [CHANGELOG](https://github.com/kowainik/summoner/blob/master/CHANGELOG.md) with the short description of my latest changes.
- [X] All new and existing tests pass.
- [X] I keep the code style used in the files I've changed (see [style-guide](https://github.com/kowainik/org/blob/master/style-guide.md#haskell-style-guide) for more details).
- [ ] I've used the [`stylish-haskell` file](https://github.com/kowainik/summoner/blob/master/.stylish-haskell.yaml).
- [X] I changed code related to the generated project.
  - [X] I created a new project using `summoner` and I checked that this specific feature is working as expected.


.travis.yml
```diff
@@ -34,6 +34,17 @@ matrix:
         - ghc-8.4.3
         - cabal-install-head
 
+  - ghc: 8.4.4
+    env: GHCVER='8.4.3' CABALVER='head'
```
vrom911:
```suggestion
    env: GHCVER='8.4.4' CABALVER='head'
```

.travis.yml
```diff
@@ -34,6 +34,17 @@ matrix:
...
+    os: linux
+    addons:
+      apt:
+        sources:
+        - hvr-ghc
+        packages:
+        - ghc-8.4.3
```
vrom911:
```suggestion
        - ghc-8.4.4
```

src/Summoner/GhcVer.hs
```diff
@@ -44,3 +47,4 @@ baseVer = \case
     Ghc802  -> "4.9.1.0"
     Ghc822  -> "4.10.1.0"
     Ghc843  -> "4.11.1.0"
+    Ghc844  -> "4.12.0.0"
```
vrom911: According to [this release notes] the `base` version should be `4.11.1.0`
```suggestion
    Ghc844  -> "4.11.1.0"
```

cronokirby:
I think making GHC 8.4.4 the default is a good idea.
I tested this change by generating a project, and it suggested GHC 8.4.4; browsing the code, the only place I saw generating this default option was looking at the latest version in https://github.com/kowainik/summoner/blob/master/src/Summoner/GhcVer.hs through `maxBound`. Is there another place where the default GHC version is defined?

vrom911:
@cronokirby, I believe that at the moment default stack file is generated for `8.4.3` as well as Travis because of the currently implemented logic.

The default version is defined here and as you mentioned its `maxBound`:

https://github.com/kowainik/summoner/blob/f5b12ed01e3f5c607398f443153f4f5746557c50/src/Summoner/Default.hs#L20-L22

But in the code itself you'll see places with pattern matching on `Ghc843` as on default one:

https://github.com/kowainik/summoner/blob/f5b12ed01e3f5c607398f443153f4f5746557c50/src/Summoner/Template/Stack.hs#L24

But it also should be changed in PR #194 which @StevenXL is currently working at.

@ChShersh , I'm voting for removing `8.4.3` check at CI 👍

ChShersh:
Also, we should remember to update this section in the `.cabal` file:

* https://github.com/kowainik/summoner/blob/f5b12ed01e3f5c607398f443153f4f5746557c50/summoner.cabal#L19-L21

ChShersh:
@cronokirby Let's not delete `GHC-8.4.3` for now. See comment here:

* https://github.com/kowainik/summoner/pull/194#issuecomment-436565704

cronokirby:
I think this should be it for now....

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@
 * [#168](https://github.com/kowainik/summoner/issues/168):
   Upgrade `stack` `intall-cabal` option's version to `2.2.0.1` in
   the generated Travis file.
+* [#197](https://github.com/kowainik/summoner/issues/197)
+  Build with `stack` lts 12.17, and support `GHC` 8.4.4 in
```
vrom911: I guess we should explicitly say that the default is changed so it's kinda breaking, but this can be done later 🙂

PR #217, opened 2018-11-20
Add textField for Forms

vrom911:
🔥 🚒 🔥 🚒 🔥 

Ctr+d now deletes the whole input in owner field.
Changed colours just to play
Ctrl+Enter doesn't work

Resolves 💢

### DIFF
--- a/src/Summoner/Tui.hs
+++ b/src/Summoner/Tui.hs
@@ -10,7 +10,8 @@ module Summoner.Tui
        ) where
 
 import Brick (App (..), AttrMap, BrickEvent (VtyEvent), Padding (Pad), Widget, attrMap, continue,
-              customMain, hBox, halt, padRight, padTop, txt, vBox, vLimit, withAttr, (<+>), (<=>))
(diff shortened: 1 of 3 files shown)


src/Summoner/Tui.hs
```diff
@@ -112,16 +113,16 @@ mkForm = setFormConcat arrangeColumns . newForm
         ]
     )
   where
-    label :: String -> Widget n -> Widget n
-    label s w = str s <+>  w
+    label :: Text -> Widget n -> Widget n
+    label l w = txt l <+>  w
```
ChShersh: Heh

screenshot from 2018-11-20 23-10-05

vrom911: Oh no 😢

willbasky:
My general suggestion is to make parameters longer and semantic.

vrom911:
If you're talking about the var names it's mostly because it's copy pasted from sources. But we will do cleaning later along with more docs + common functions. Just need to make the approximate picture and to prettify it later 🙂

willbasky:
@vrom911 then approved!

Issue #208, opened 2018-11-18
Fundamental TUI design


vrom911:
Implement the following sections in the form:

+ [x] User (by @ChShersh)
    + [x] Owner: [input field]
    + [x] Full name: [input field] 
    + [x] E-mail: [input field] 
+ [x] Tools (✅ Cabal ✅ Stack) (by @vrom911)
+ [x] Project (by @vrom911)
    + [x] Name: [input field] !On-fly checking of the uniqueness!
    + [x] Description:  [input field]
    + [x] Category: [input field]
    + [x] License: <scrolling list>
+ [x] Project meta (by @vrom911)
   + [x] ✅ Library
   + [x] ✅ Exe
   + [x] ✅ Test
   + [x] ✅ Bench
   + [x] GHC: [space-separated versions]
   + [x] Custom Prelude [input field]
        + [x] Custom Prelude module name [input field]
+ [x] GitHub (by @ChShersh)
    + [x] Enable (⏺ Yes ⏺ No)
    + [x] ✅ Private
    + [x] ✅ Travis
    + [x] ✅ AppVeyor
+ [x] On-the-fly window with project structure (by @vrom911)
+ [x] Box with all error messages (by @ChShersh)
+ [x] Help window: edit commands, read color meaning?... (by @ChShersh )

Tech debt:
+ [x] Horizontal checkbox arrangement (by @vrom911)
+ [x] Edit box label alignment (by @vrom911)
+ [x] Padding between input fields (by @vrom911)
+ [x] Project name validation (by @ChShersh)
+ [x] Padding ratio button (by @ChShersh)
+ [x] Learn how to insert text labels in forms (by @vrom911)
+ [x] Cancel on <kbd>Esc</kbd> and submit on <kbd>Enter</kbd> (by @vrom911)
+ [x] Create project (by @vrom911)
+ [x] Disabled checkboxes (by @ChShersh)
+ [x] Focus ring check to delete every input field (by @vrom911 and @ChShersh)
+ [x] Improve `help` (by @vrom911)
+ [x] Delete input field content by <kbd>Ctrl</kbd>+<kbd>d</kbd> (by @vrom911)
+ [x] Border label padding (by @ChShersh)
+ [x] Remove bold directory (by @vrom911)
+ [x] Pad project structure to the bottom (by @vrom911)
+ [x] No upload (by @ChShersh)
+ [x] Wire up CLI interface (by @ChShersh)
+ [x] Read config on start (by @vrom911)
[...]

PR #218, opened 2018-11-20
[#208] Make checkboxes disableable

ChShersh:
Resolves :rage1: 

Required to copy-paste the implementation of `customCheckboxField` and refactor it.

## Checklist:

- [ ] I've updated the [CHANGELOG](https://github.com/kowainik/summoner/blob/master/CHANGELOG.md) with the short description of my latest changes.
- [ ] All new and existing tests pass.
- [ ] I keep the code style used in the files I've changed (see [style-guide](https://github.com/kowainik/org/blob/master/style-guide.md#haskell-style-guide) for more details).
- [ ] I've used the [`stylish-haskell` file](https://github.com/kowainik/summoner/blob/master/.stylish-haskell.yaml).

### DIFF
--- a/src/Summoner/Tui.hs
+++ b/src/Summoner/Tui.hs
@@ -9,7 +9,7 @@ module Summoner.Tui
        ( summonTui
        ) where
 
-import Brick (App (..), AttrMap, BrickEvent (VtyEvent), Padding (Pad), Widget, attrMap, continue,
+import Brick (App (..), AttrMap, BrickEvent (..), Padding (Pad), Widget, attrMap, continue,
               customMain, hBox, halt, padRight, padTop, str, txt, vBox, vLimit, withAttr, (<+>),
               (<=>))
 import Brick.Focus (focusGetCurrent, focusRingCursor)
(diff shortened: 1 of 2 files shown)


src/Summoner/Tui.hs
```diff
@@ -75,7 +77,7 @@ data SummonForm
 
 -- Creates the inout form from the given initial 'TreasureChest'.
 mkForm :: forall e . SummonKit -> Form SummonKit e SummonForm
-mkForm = setFormConcat arrangeColumns . newForm
+mkForm sk = setFormConcat arrangeColumns $ newForm
```
willbasky: may be `sKit` or `skit` will be better.

src/Summoner/Tui.hs
```diff
@@ -173,6 +179,10 @@ app = App
         VtyEvent V.EvResize {}       -> continue s
```
willbasky: `s`?

src/Summoner/Tui.hs
```diff
@@ -173,6 +179,10 @@ app = App
...
         -- TODO: should we cancel on Esc and apply on Ctrl+Enter?
         VtyEvent (V.EvKey V.KEsc []) -> halt s
+        MouseDown n _ _ _ -> case n of
+            GitHubEnable  -> handleFormEvent ev s >>= continue . mkForm . formState
+            GitHubDisable -> handleFormEvent ev s >>= continue . mkForm . formState
+            _             -> handleFormEvent ev s >>= continue
```
willbasky: `ev`?

src/Summoner/Tui.hs
ChShersh: `s` has type `Form SummonKit e SummonForm`. Maybe `form` is better name...

src/Summoner/Tui.hs
ChShersh: `ev` is for event. This probably will be refactored later.

src/Summoner/Tui.hs
willbasky: `formState` or `formApp`?

src/Summoner/Tui.hs
willbasky: `event` if nobody uses it/

PR #221, opened 2018-11-21
Add basic structured tree widget

vrom911:
Resolves :christmas_tree:

### DIFF
--- a/src/Summoner/Tui.hs
+++ b/src/Summoner/Tui.hs
@@ -29,6 +29,7 @@ import qualified Brick (on)
 import qualified Brick.Util as U
 import qualified Brick.Widgets.Border as B
 import qualified Brick.Widgets.Center as C
+import qualified Brick.Widgets.Core as W
 import qualified Brick.Widgets.Edit as E
 import qualified Brick.Widgets.List as L
 import qualified Data.Text as T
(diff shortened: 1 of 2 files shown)


src/Summoner/Tui.hs
```diff
@@ -156,7 +155,8 @@ draw :: Form SummonKit e SummonForm -> [Widget SummonForm]
 draw f =
     [ C.vCenter $ vBox
         [ C.hCenter form
-        , validationErrors $ invalidFields f
+      <+> W.hLimitPercent 25 (C.hCenter tree)
+        , validationErrors (invalidFields f)
```
ChShersh: That's nice! I like new arrangement very much. Also it was unexpected that it works with just putting text there.

src/Summoner/Tui/Kit.hs
```diff
@@ -50,52 +51,55 @@ module Summoner.Tui.Kit
...
 
 -- | Global TUI state.
 data SummonKit = SummonKit
-    { summonKitUser        :: User
-    , summonKitProject     :: Project
-    , summonKitCabal       :: Bool
-    , summonKitStack       :: Bool
-    , summonKitProjectMeta :: ProjectMeta
-    , summonKitGitHub      :: GitHub
+    { summonKitUser        :: !User
```
ChShersh: Thank you for adding strict fields :+1:

PR #224, opened 2018-11-22
Move out Form and Validation logic

vrom911:
Resolves 👾

### DIFF
--- a/src/Summoner/Tui.hs
+++ b/src/Summoner/Tui.hs
@@ -128,17 +128,19 @@ app dirs = App
         VtyEvent (V.EvKey V.KEnter []) -> halt s
         VtyEvent (V.EvKey V.KEsc []) -> halt s
         VtyEvent (V.EvKey (V.KChar 'd') [V.MCtrl]) ->
-            handleFormEvent ev s >>= continue . summonFormValidation dirs . ctrlD
+            withForm ev s (summonFormValidation dirs . ctrlD)
         MouseDown n _ _ _ -> case n of
-            GitHubEnable  -> handleFormEvent ev s >>= continue . mkForm . formState
-            GitHubDisable -> handleFormEvent ev s >>= continue . mkForm . formState
-            _             -> handleFormEvent ev s >>= continue
-        _ -> handleFormEvent ev s >>= continue . summonFormValidation dirs
+            GitHubEnable  -> withForm ev s (mkForm . formState)
+            GitHubDisable -> withForm ev s (mkForm . formState)
+            _             -> withForm ev s id
+        _ -> withForm ev s (summonFormValidation dirs)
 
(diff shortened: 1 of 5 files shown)


src/Summoner/Tui.hs
```diff
@@ -200,53 +106,19 @@ app :: [FilePath] -> App (Form SummonKit e SummonForm) e SummonForm
...
-                if focusGetCurrent (formFocus s') == Just UserOwner
-                then mkForm $ formState s' & user . owner .~ ""
-                else s'
-
+        VtyEvent (V.EvKey (V.KChar 'd') [V.MCtrl]) ->
+            handleFormEvent ev s >>= continue . summonFormValidation dirs . ctrlD
         MouseDown n _ _ _ -> case n of
             GitHubEnable  -> handleFormEvent ev s >>= continue . mkForm . formState
             GitHubDisable -> handleFormEvent ev s >>= continue . mkForm . formState
             _             -> handleFormEvent ev s >>= continue
```
ChShersh: The pattern with `handleFormEvent ev s >>= continue . f` appears in many places. Now, after refactoring fields validation into separate function it's more visually clear. Maybe it makes sense to implement some function like:

```haskell
withForm ev s f = handleFormEvent ev s >>= continue . f
```

What do you think? Doesn't really decrease the number of lines, but maybe makes existing lines shorter...

src/Summoner/Tui/Validation.hs
```diff
@@ -0,0 +1,68 @@
...
+      clearField UserFullName (user . fullName)
+    . clearField UserEmail (user . email)
+    . clearField ProjectName (project . repo)
+    . clearField ProjectDesc (project . desc)
+    . clearField ProjectCat (project . category)
+    . clearField CustomPreludeName (projectMeta . preludeName)
+    . clearField CustomPreludeModule (projectMeta . preludeModule)
+    . clearField UserOwner (user . owner)
+  where
+    clearField :: SummonForm -> Lens' SummonKit Text -> KitForm e -> KitForm e
```
ChShersh: After thinking about the way to clean other fields rather than `Text` I have one idea. You can implement function where you pass value to set with lens, like:

```haskell
clearField :: a -> SummonForm -> Lens' SummonKit a -> KitForm e -> KitForm e
```

And then you can have:

```haskell
clearTextField :: SummonForm -> Lens' SummonKit Text -> KitForm e -> KitForm e
clearTextField = clearField ""

clearListField :: SummonForm -> Lens' SummonKit [a] -> KitForm e -> KitForm e
clearListField = clearField []
```

Or you can pass `""` or `[]` directly as arguments (looks like `clearField ""` is much shorter than `clearTextField`). Not sure it worth the complication but maybe useful if we want to be able to clear all fields.

PR #227, opened 2018-11-23
[#208] Fill SummonKit by the config and CLI

vrom911:
Resolves 🍶


src/Summoner/Tui/Kit.hs
```diff
@@ -104,40 +110,6 @@ data GitHub = GitHub
     , gitHubAppVeyor :: !Bool
     } deriving (Show)
 
--- | Initial global state of the tui.
-initialSummonKit :: SummonKit
```
ChShersh: Aaaaaand it's over :video_game:

PR #238, opened 2018-11-24
[#208] Move summon-tui into separate package

ChShersh:
Resolves :package:


summoner-tui/summoner-tui.cabal
```diff
@@ -0,0 +1,88 @@
...
+homepage:            https://github.com/kowainik/summoner
+bug-reports:         https://github.com/kowainik/summoner/issues
+license:             MPL-2.0
+license-file:        LICENSE
+author:              Kowainik
+maintainer:          xrom.xkov@gmail.com
+copyright:           2018 Kowainik
+category:            TUI, TUI Tool, Development
+build-type:          Simple
+stability:           stable
```
vrom911: *coughing* 
heh

summoner-tui/summoner-tui.cabal
```diff
@@ -0,0 +1,88 @@
...
+  exposed-modules:     Summoner.Tui
+                         Summoner.Tui.Field
+                         Summoner.Tui.Form
+                         Summoner.Tui.GroupBorder
+                         Summoner.Tui.Kit
+                         Summoner.Tui.Validation
+                         Summoner.Tui.Widget
+
+  autogen-modules:     Paths_summoner_tui
+  other-modules:       Paths_summoner_tui
```
vrom911: I now think that we should somehow use this in CLI `--version` option. As we are having different versions for `summoner` and `summoner-tui` we should reflect it in this command as well. Do you think it's doable?

PR #241, opened 2018-11-25
[#208] Add confirm window on Enter

vrom911:
Resolves 🌬


summoner-tui/src/Summoner/Tui.hs
```diff
@@ -112,13 +123,21 @@ appNew dirs = App
 
 -- | Draws the form for @new@ command.
 drawNew :: [FilePath] -> KitForm e -> [Widget SummonForm]
-drawNew dirs f =
-    [ vBox
-        [ form <+> tree
-        , validationErrors <+> help
+drawNew dirs f = case formState f ^. shouldSummon of
+    Idk -> [confirm]
+    _   ->
```
ChShersh: What do you think about writing on top-level in this case more high-overview structure, like:

```haskell
    Idk -> [confirmDialog]
    _   -> [formWidget]
```

Issue #275, opened 2019-01-03
Change Stack installation at CI


vrom911:
Stack installation can use the official script instead of downloading the tarball and unpacking it, like this:

```bash
  - curl -sSL https://get.haskellstack.org/ | sh
```

 + [x] Summoner
 + [ ] Generated project

Issue #272, opened 2018-12-11
Simplify Travis config for cabal


ChShersh:
Travis CI config for `cabal` can be much simpler than it's now. See example here:

* https://github.com/haskell/cabal/issues/5787#issuecomment-446144752

- [x] Summoner
- [ ] Generated project

PR #276, opened 2019-01-03
[#275] [#272] New simple travis settings

vrom911:
Implements `summoner` part of the issues #275 and #272. The same improvements for the generated projects still remain.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,9 +40,10 @@ matrix:
     env: GHCVER='8.6.3' STACK_YAML="$TRAVIS_BUILD_DIR/stack.yaml" DEPLOY=yes
     os: linux
 
-  # macOS build
-  - env: GHCVER='8.6.3' STACK_YAML="$TRAVIS_BUILD_DIR/stack.yaml" DEPLOY=yes
-    os: osx
+#  Disabled because of Travis bugs
+#  # macOS build
+#  - env: GHCVER='8.6.3' STACK_YAML="$TRAVIS_BUILD_DIR/stack.yaml" DEPLOY=yes
+#    os: osx
 
 install:
   - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ cache:
 
 matrix:
   include:
+  # Cabal
   - ghc: 8.2.2
     env: GHCVER='8.2.2' CABALVER='head'
     os: linux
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,82 +1,50 @@
 sudo: true
-language: generic
+language: haskell
```
ChShersh: Let's also add one line here with the following content:

```yaml
dist: trusty
```

I guess this allows to fetch GHC versions from hvr PPA

.travis.yml
```diff
@@ -1,82 +1,50 @@
...
 
 git:
   depth: 5
 
+cabal: "head"
```
ChShersh: Interesting idea! I guess it makes sense to build `summoner` with `head` version of `cabal`. But in scaffolded projects it's probably better to use specific version of cabal, that will be in `.cabal` file as well.

.travis.yml
```diff
@@ -1,82 +1,50 @@
...
+
 cache:
   directories:
   - "$HOME/.cabal"
+  - "$HOME/.cabal/store"
```
ChShersh: I think `.cabal/store` can be used instead of `.cabal`. Looks like it's enough to cache only `store` subdirectory of `.cabal` folder. I'm not aware of full content for `.cabal` directory though...

.travis.yml
```diff
@@ -1,82 +1,50 @@
...
   - "$HOME/.stack"
   - "$TRAVIS_BUILD_DIR/.stack-work"
 
 matrix:
   include:
+  # Cabal
   - ghc: 8.2.2
     env: GHCVER='8.2.2' CABALVER='head'
     os: linux
-    addons:
```
ChShersh: Nice cleanup!

.travis.yml
```diff
@@ -86,16 +54,8 @@ install:
...
-      export PATH="$HOME/.local/bin:$PATH"
-
       # install stack
-      export STACK_URL="https://www.stackage.org/stack/${TRAVIS_OS_NAME}-x86_64"
-      echo "Download stack from: $STACK_URL"
-      cd /tmp
-      curl --location "$STACK_URL" > stack.tar.gz
-      tar xz --strip-components=1 --file=stack.tar.gz -C ~/.local/bin/
-      cd -
+      curl -sSL https://get.haskellstack.org/ | sh
```
ChShersh: This is very nice :+1:

.travis.yml
```diff
@@ -86,16 +54,8 @@ install:
       cabal new-update
```
ChShersh: I wonder if exporting `$PATH` variable is still required with new settings for cabal and ghc? :thinking:

.travis.yml
```diff
@@ -1,82 +1,50 @@
...
 cache:
   directories:
   - "$HOME/.cabal"
+  - "$HOME/.cabal/store"
   - "$HOME/.stack"
   - "$TRAVIS_BUILD_DIR/.stack-work"
 
 matrix:
   include:
+  # Cabal
```
ChShersh: Btw, I like these comments!

.travis.yml
vrom911: Nice catch!

PR #279, opened 2019-01-04
[#272] [#275] Travis improvements for generated project

### LINKED ISSUE #272: Simplify Travis config for cabal

### LINKED ISSUE #275: Change Stack installation at CI

vrom911:
It doesn't work for Stack when the version of GHC is 8.0.2 because it uses old Stack, not the one that is configured, but it shouldn't work with the old version of Cabal (manually writing `stack setup --no-terminal --install-cabal 2.2.0.1` doesn't help as well.

__UPD:__ We decided not to add old GHC into Stack Travis matrix.

Resolves #272 
Resolves #275 

## Checklist:

* Update CHANGELOG

### DIFF
--- a/summoner-cli/src/Summoner/Default.hs
+++ b/summoner-cli/src/Summoner/Default.hs
@@ -2,6 +2,7 @@
 
 module Summoner.Default
        ( defaultGHC
+       , defaultCabal
        , defaultTomlFile
        , defaultConfigFile
        , defaultDescription
@@ -21,6 +22,10 @@ import Summoner.GhcVer (GhcVer)
 -- | Default GHC version is the latest available.
 defaultGHC :: GhcVer
 defaultGHC = maxBound
+
+-- | Default version of the Cabal.
+defaultCabal :: Text
+defaultCabal = "2.0"
 
 defaultTomlFile :: FilePath
 defaultTomlFile = ".summoner.toml"
--- a/summoner-cli/src/Summoner/GhcVer.hs
+++ b/summoner-cli/src/Summoner/GhcVer.hs
@@ -1,3 +1,7 @@
+{- | Contains data type for GHC versions supported by Summoner
+and some useful functions for manipulation with them.
+-}
+
 module Summoner.GhcVer
        ( GhcVer (..)
        , Pvp (..)
@@ -6,6 +10,8 @@ module Summoner.GhcVer
        , latestLts
        , baseVer
        , cabalBaseVersions
+
+       , oldGhcs
        ) where
 
 import Data.List (maximum, minimum)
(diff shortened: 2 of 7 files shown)


summoner-cli/src/Summoner/Default.hs
```diff
@@ -22,6 +23,10 @@ import Summoner.GhcVer (GhcVer)
 defaultGHC :: GhcVer
 defaultGHC = maxBound
 
+-- | Default version of the Cabal.
+defaultCabal :: Text
```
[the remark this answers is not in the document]
vrom911: Yes, because our plans are to make it more flexible in the future :)

vrom911:
@ChShersh I don't want to make not working settings, at least we shouldn't include this GHC version in Stack matrix and specify this behaviour explicitly somewhere with the reasoning. I will try to ask some people; probably somebody already solved this problem.

ChShersh:
@vrom911 This is a very reasonable approach :+1: I agree, let's first ask, maybe something can be done with this. Otherwise - I have no idea how to fix this problem...

vrom911:
Not sure it's the best place but:

https://twitter.com/vronnie911/status/1081317117977378816

(it's not really can be an issue to Stack, as they would say that this is feature)

summoner-cli/src/Summoner/GhcVer.hs
```diff
@@ -35,6 +41,10 @@ showGhcVer = \case
     Ghc844  -> "8.4.4"
     Ghc863  -> "8.6.3"
 
+-- | These are old GHC versions that are not working with default GHC versions when using Stack.
+oldGhcs :: [GhcVer]
+oldGhcs = [Ghc7103, Ghc801, Ghc802]
```
ChShersh: Can this be written as `[minBound .. Ghc802]`? I assume that once we bump up cabal-version, we only need to bump up upper bound in this list. I guess that the problem with cabal doesn't work for all GHC versions in sequence up to some version..

summoner-cli/src/Summoner/Template/GitHub.hs
```diff
@@ -1,15 +1,29 @@
 {-# LANGUAGE QuasiQuotes  #-}
 {-# LANGUAGE ViewPatterns #-}
 
+{-| This module contains template for GitHub related docs:
```
ChShersh: This documentation is soooo niiice! :clap:

Issue #150, opened 2018-09-18
Add `.nix` file to `summoner` as well


chshersh:
I assume it's content should be consistent with issue #149

mgttlinger:
Given that my current workflow of copying and modifying my nix files for each new Haskell project is rather cumbersome I would be interested in working on this issue. 

Where should I look in the source to get up to speed and what exactly shall be generated?

ChShersh:
@mgttlinger This issue is about adding `.nix` file to the `summoner` GitHub repo and, preferably, patch `.travis.yml` to build with `nix` on CI and add instructions how to build with `.nix` or what should be done to update configuration and when `.nix` configuiration requires update. For teaching `summoner` to generate `.nix` files, see issue #149

mgttlinger:
Oh, then I misunderstood. But I#d be interested in doing both.

chessai:
I have something that generates a nix config for a project that works, will have a PR shortly

chessai:
i imagine the PR will need some review, i'm very unfamiliar with travis

ChShersh:
@chessai This looks very nice! And it would very kind of you to give some short description to every file and why it's needed :slightly_smiling_face: I would like `summoner` to be beginner-friendly tool, that's why description and comments are valuable! For example, how users can build with different GHC versions using `nix` and so on. I'm not using `nix`, so I'm not aware of the workflow... But if there're simple build command, this can help to modify `.travis.yml` as well to build with nix on CI and test there. So this is useful not only for beginners :smile:

chessai:
@ChShersh understood

PR #283, opened 2019-01-07
add nix and nixpkg scaffolding

chessai:
doesn't really resolve an issue, but the thing it addresses was mentioned in #150.

## Checklist:

* Update CHANGELOG
- [ ] Keep code style used in the changed files (see [style-guide](https://github.com/kowainik/org/blob/master/style-guide.md#haskell-style-guide) for more details).
- [ ] Use the [`stylish-haskell` file](https://github.com/kowainik/summoner/blob/master/.stylish-haskell.yaml).
- [ X ] Create a new test project using `summoner` and check that the changes work as expected.

### DIFF
--- a/summoner-cli/src/Summoner/Project.hs
+++ b/summoner-cli/src/Summoner/Project.hs
@@ -21,9 +21,9 @@ import Summoner.License (LicenseName (..), customizeLicense, fetchLicense, licen
                          parseLicenseName)
 import Summoner.Process ()
 import Summoner.Question (YesNoPrompt (..), checkUniqueName, choose, falseMessage,
-                          mkDefaultYesNoPrompt, query, queryDef, queryManyRepeatOnFail,
-                          targetMessageWithText, trueMessage)
-import Summoner.Settings (CustomPrelude (..), Settings (..))
+                          mkDefaultYesNoPrompt, query, queryNotNull, queryDef, queryManyRepeatOnFail,
+                          targetMessageWithText, trueMessage, chooseYesNo)
+import Summoner.Settings (CustomPrelude (..), Settings (..), NixPkgSet (..), defaultNixPkgSet, showNixPkgSet)
 import Summoner.Source (fetchSource)
 import Summoner.Template (createProjectTemplate)
 import Summoner.Text (intercalateMap, packageToModule)
@@ -40,7 +40,8 @@ generateProject
 generateProject settingsNoUpload isOffline projectName Config{..} = do
     settingsRepo <- checkUniqueName projectName
     -- decide cabal stack or both
-    (settingsCabal, settingsStack) <- getCabalStack (cCabal, cStack)
+    (settingsCabal, settingsStack, settingsNix) <- getCabalStackNix (cCabal, cStack, cNix)
+    settingsNixPkgSet <- getNixPkgSet settingsNix
 
     settingsOwner       <- queryDef "Repository owner: " cOwner
     settingsDescription <- queryDef "Short project description: " defaultDescription
@@ -154,26 +155,62 @@ generateProject settingsNoUpload isOffline projectName Config{..} = do
                 pure $ Just $ CustomPrelude p m
         Last prelude@(Just (CustomPrelude p _)) ->
             prelude <$ successMessage ("Custom prelude " <> p <> " will be used in the project")
-
+    
+    getNixPkgSet :: Bool -> IO (Maybe NixPkgSet)
+    getNixPkgSet usingNix = if usingNix
+      then case cNixPkgSet of
+        Last Nothing -> do
+          let yesDo, noDo :: IO (Maybe NixPkgSet)
+              yesDo = do
+                o <- queryNotNull "Nix package set GitHub owner name: "
+                r <- queryNotNull "Nix package set GitHub repo name: "
+                rv <- queryNotNull "Nix package set Git revision: "
+                s  <- queryNotNull "Nix package set unpacked SHA256: "
+                let pkgSet = NixPkgSet { npsOwner = o, npsRepo = r, npsRev = rv, npsSha = s }
+                successMessage ("The nix package set " <> showNixPkgSet pkgSet <> " will be used in the project")
+                pure $ Just pkgSet
+              noDo = do
+                successMessage ("The default nix package set " <> showNixPkgSet defaultNixPkgSet <> " will be used in the project")
+                pure $ Just defaultNixPkgSet 
+          chooseYesNo (mkDefaultYesNoPrompt "nix package set") yesDo noDo
+        Last set@(Just n) -> set <$ successMessage ("The nix package set " <> showNixPkgSet n <> " will be used in the project")
+      else pure Nothing
+    
     -- get what build tool to use in the project
     -- If user chose only one during CLI, we assume to use only that one.
-    getCabalStack :: (Decision, Decision) -> IO (Bool, Bool)
-    getCabalStack = \case
-        (Idk, Idk) -> decisionToBool cCabal (mkDefaultYesNoPrompt "cabal") >>= \c ->
-            if c then decisionToBool cStack (mkDefaultYesNoPrompt "stack") >>= \s -> pure (c, s)
-            else stackMsg True >> pure (False, True)
-        (Nop, Nop) -> errorMessage "Neither cabal nor stack was chosen" >> exitFailure
-        (Yes, Yes) -> output (True, True)
-        (Yes, _)   -> output (True, False)
-        (_, Yes)   -> output (False, True)
-        (Nop, Idk) -> output (False, True)
-        (Idk, Nop) -> output (True, False)
+    getCabalStackNix :: (Decision, Decision, Decision) -> IO (Bool, Bool, Bool)
+    getCabalStackNix = \case
+        (Idk, Idk, Idk) -> do
+          c <- decisionToBool cCabal (mkDefaultYesNoPrompt "cabal")
+          s <- decisionToBool cStack (mkDefaultYesNoPrompt "stack")
+          n <- decisionToBool cNix (mkDefaultYesNoPrompt "nix")
+          when (not c && n) nixWithoutCabalError
+          output (c, s, n)
+        (Nop, Nop, Nop) -> noneOfError
+        (Yes, Yes, Yes) -> output (True, True, True)
+        (Idk, Idk, Nop) -> output (True, False, False)
+        (Idk, Idk, Yes) -> output (True, False, True)
+        (Idk, Nop, _)   -> output (True, False, False)
+        (Idk, Yes, _)   -> output (True, True, False)
+        (Nop, Idk, _)   -> output (False, True, False)
+        (Nop, Nop, Idk) -> noneOfError
+        (Nop, Nop, Yes) -> nixWithoutCabalError
+        (Nop, Yes, _)   -> output (False, True, False)
+        (Yes, Idk, _)   -> output (True, False, False)
+        (Yes, Nop, _)   -> output (True, False, False)
+        (Yes, Yes, Idk) -> output (True, True, False)
+        (Yes, Yes, Nop) -> output (True, True, False)
       where
-        output :: (Bool, Bool) -> IO (Bool, Bool)
-        output x@(c, s) = cabalMsg c >> stackMsg s >> pure x
+        output :: (Bool, Bool, Bool) -> IO (Bool, Bool, Bool)
+        output x@(c, s, n) = cabalMsg c >> stackMsg s >> nixMsg n >> pure x
 
         cabalMsg c = targetMessageWithText c "Cabal" "used in this project"
         stackMsg c = targetMessageWithText c "Stack" "used in this project"
+        nixMsg   c = targetMessageWithText c "Nix"   "used in this project"
+
+        nixWithoutCabalError = errorMessage "Using nix without cabal is an unsupported configuration" >> exitFailure
+        noneOfError = errorMessage "None of {cabal,stack,nix} was chosen" >> exitFailure
+
 
 -- | Creates the directory and run GitHub commands.
 initializeProject :: Settings -> IO ()
--- a/summoner-cli/src/Summoner/Settings.hs
+++ b/summoner-cli/src/Summoner/Settings.hs
@@ -2,6 +2,10 @@ module Summoner.Settings
        ( Settings (..)
        , CustomPrelude (..)
        , customPreludeT
+       , NixPkgSet(..)
+       , nixPkgSetT
+       , defaultNixPkgSet
+       , showNixPkgSet
        ) where
 
 import Toml (TomlCodec, (.=))
@@ -10,7 +14,6 @@ import Summoner.GhcVer (GhcVer)
 import Summoner.License (License, LicenseName)
 
 import qualified Toml
-
 
 data CustomPrelude = CustomPrelude
     { cpPackage :: Text
@@ -21,6 +24,42 @@ customPreludeT :: TomlCodec CustomPrelude
 customPreludeT = CustomPrelude
     <$> Toml.text "package" .= cpPackage
     <*> Toml.text "module"  .= cpModule
+
+data NixPkgSet = NixPkgSet
+    { npsOwner :: Text
+    , npsRepo  :: Text
+    , npsRev   :: Text
+    , npsSha   :: Text
+    } deriving (Show, Eq)
+
+nixPkgSetT :: TomlCodec NixPkgSet
+nixPkgSetT = NixPkgSet
+    <$> Toml.text "owner"  .= npsOwner
+    <*> Toml.text "repo"   .= npsRepo
+    <*> Toml.text "rev"    .= npsRev
+    <*> Toml.text "sha256" .= npsSha
+
+defaultNixPkgSet :: NixPkgSet
+defaultNixPkgSet = NixPkgSet
+  { npsOwner = "NixOS" 
+  , npsRepo  = "nixpkgs" 
+  , npsRev   = "cecec1f74468766825c2ad32d8388c2ded36225f"
+  , npsSha   = "1sq538wy0shbakah27b6n4bl5amzwkzjsds77vdd8rsq0d1nys4w"
+  }
+
+showNixPkgSet :: NixPkgSet -> Text                                                                   
+showNixPkgSet NixPkgSet{..} = mconcat                                                                
+  [ "https://github.com/"
+  , npsOwner 
+  , "/"        
+  , npsRepo 
+  , "/archive/"
+  , npsRev
+  , ".tar.gz"
+  , " (SHA256 is "
+  , npsSha
+  , ")" 
+  ]
 
 -- | Data needed for project creation.
 data Settings = Settings
@@ -48,6 +87,8 @@ data Settings = Settings
     , settingsWarnings       :: ![Text] -- ^ default warnings
     , settingsCabal          :: !Bool
     , settingsStack          :: !Bool
+    , settingsNix            :: !Bool
+    , settingsNixPkgSet      :: !(Maybe NixPkgSet)
     , settingsStylish        :: !(Maybe Text) -- ^ @.stylish-haskell.yaml@ file
     , settingsContributing   :: !(Maybe Text) -- ^ @CONTRIBUTING.md@ file
     , settingsNoUpload       :: !Bool  -- ^ do not upload to GitHub
--- a/summoner-cli/src/Summoner/Template/Nix.hs
+++ b/summoner-cli/src/Summoner/Template/Nix.hs
@@ -1,0 +1,142 @@
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE NamedFieldPuns #-}
+
+module Summoner.Template.Nix
+       ( nixFiles
+       ) where
+
+import NeatInterpolation (text)
+
+import Summoner.Default (defaultGHC)
+import Summoner.GhcVer (GhcVer (..))
+import Summoner.Settings (Settings (..), NixPkgSet (..), defaultNixPkgSet)
+import Summoner.Tree (TreeFs (..))
+
+import qualified Data.Text as T
+
+nixCompiler :: GhcVer -> Text
+nixCompiler = \case
+  Ghc7103 -> "ghc7103"
+  Ghc801  -> "ghc802"
+  Ghc802  -> "ghc802"
+  Ghc822  -> "ghc822"
+  Ghc843  -> "ghc844" 
+  Ghc844  -> "ghc844"
+  Ghc863  -> "ghc863"
+
+defaultNixCompiler :: Text
+defaultNixCompiler = nixCompiler defaultGHC
+
+nixFiles :: Settings -> [TreeFs]
+nixFiles Settings{..} =
+  [ shellNix
+  , releaseNix
+  , defaultNix
+  , nixDir
+  ]
+  where
+    defaultNix = File "default.nix" defaultNixT
+    shellNix = File "shell.nix" shellNixT
+    releaseNix = File "release.nix" releaseNixT
+
+    nixDir = Dir "nix" [fetchNixpkgsNix, nixpkgsNix, overridesNix]
+    fetchNixpkgsNix = File "fetchNixpkgs.nix" fetchNixpkgsNixT
+    nixpkgsNix = File "nixpkgs.nix" nixpkgsNixT
+    overridesNix = File "overrides.nix" overridesNixT
+
+    defaultNixT :: Text
+    defaultNixT =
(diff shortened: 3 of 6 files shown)


chessai:
noting so i don't forget: i want to add a feature allowing users to specify how nix will clean the source (ie stuff like .ghc or dist* files won't affect builds)

also, the output when you say 'no' to using a custom nixpkg set is slightly wrong when you have nix support enabled. unpinned nixpkg sets are not allowed, so it has a default nixpkg set it falls back on. the stdout reads something like "a nixpkg set wont be added", and then the next line goes on to describe which nixpkg set you are using. i need to fix this

chessai:
also, nix has some requirements regarding the sha256. can't recall what they are, but it could be validated.

chessai:
one enhancement for nix users, if they have nix-prefetch-url on their path, is that we could fetch the nixpkgs sha256 with just the revision, so users don't have to use nix-prefetch-url themselves. i wrote a simple tool that does this [here](https://github.com/chessai/grab)

chessai:
Not sure what further documentation is needed. perhaps an example toml file?

chessai:
i think, for the clean source thing, i'd prefer to just use https://github.com/siers/nix-gitignore. it requires that the project have a .gitignore file (it uses the .gitignore to inform how it cleans the source). or there could be a `nixCleanSource = ["ignoreThis", "ignoreThis"]`. unsure. i think that should be a separate PR from this, as its not integral.

summoner-cli/src/Summoner/Project.hs
```diff
@@ -155,25 +157,61 @@ generateProject settingsNoUpload isOffline projectName Config{..} = do
         Last prelude@(Just (CustomPrelude p _)) ->
             prelude <$ successMessage ("Custom prelude " <> p <> " will be used in the project")
 
+    getNixPkgSet :: Bool -> IO (Maybe NixPkgSet)
+    getNixPkgSet usingNix = if usingNix
+        then case cNixPkgSet of
+            Last Nothing -> do
+                let yesDo, noDo :: IO (Maybe NixPkgSet)
+                    yesDo = do
+                        o <- queryNotNull "Nix package set GitHub owner name: "
```
vrom911: *Sorry, I will ask a lot of (stupid) questions, as I'm not familiar with nix.*

Is that could be any name or just limited amount of possible answers here? I noticed that in your example config you have `nixOS`, should this be the default?

summoner-cli/src/Summoner/Project.hs
```diff
@@ -155,25 +157,61 @@ generateProject settingsNoUpload isOffline projectName Config{..} = do
...
+                        r <- queryNotNull "Nix package set GitHub repo name: "
```
vrom911: The same quesion here.. And probably they are linked together so we can make it in pairs?
This doesn't mean that it should be done now, just to know for the future.

summoner-cli/src/Summoner/Project.hs
```diff
@@ -155,25 +157,61 @@ generateProject settingsNoUpload isOffline projectName Config{..} = do
...
+                        rv <- queryNotNull "Nix package set Git revision: "
+                        s  <- queryNotNull "Nix package set unpacked SHA256: "
```
vrom911: Heh, this will be tricky ones. Is there a way to somehow get this data?

summoner-cli/src/Summoner/Project.hs
```diff
@@ -155,25 +157,61 @@ generateProject settingsNoUpload isOffline projectName Config{..} = do
...
+                        let pkgSet = NixPkgSet { npsOwner = o, npsRepo = r, npsRev = rv, npsSha = s }
+                        successMessage ("The nix package set " <> showNixPkgSet pkgSet <> " will be used in the project")
+                        pure $ Just pkgSet
+                    noDo = do
+                        successMessage ("The default nix package set " <> showNixPkgSet defaultNixPkgSet <> " will be used in the project")
+                        pure $ Just defaultNixPkgSet
+                chooseYesNo (mkDefaultYesNoPrompt "nix package set") yesDo noDo
```
vrom911: Probably, add a more clear comment about what Yes and No will mean for the user in this case (that the default will be used if the user says no if I understand this correctly).

summoner-cli/src/Summoner/Project.hs
```diff
@@ -155,25 +157,61 @@ generateProject settingsNoUpload isOffline projectName Config{..} = do
...
-        output :: (Bool, Bool) -> IO (Bool, Bool)
-        output x@(c, s) = cabalMsg c >> stackMsg s >> pure x
+        output :: (Bool, Bool, Bool) -> IO (Bool, Bool, Bool)
+        output x@(c, s, n) = cabalMsg c >> stackMsg s >> nixMsg n >> pure x
 
         cabalMsg c = targetMessageWithText c "Cabal" "used in this project"
         stackMsg c = targetMessageWithText c "Stack" "used in this project"
+        nixMsg   c = targetMessageWithText c "Nix"   "used in this project"
+
+        nixWithoutCabalError = errorMessage "Using nix without cabal is an unsupported configuration" >> exitFailure
```
vrom911: Why this is the issue?
The `cabal` file should be created in any case. This mostly will affect other configs like Travis etc.

summoner-cli/src/Summoner/Project.hs
```diff
@@ -155,25 +157,61 @@ generateProject settingsNoUpload isOffline projectName Config{..} = do
...
+        (Idk, Nop, _)   -> output (True, False, False)
+        (Idk, Yes, _)   -> output (True, True, False)
+        (Nop, Idk, _)   -> output (False, True, False)
+        (Nop, Nop, Idk) -> noneOfError
+        (Nop, Nop, Yes) -> nixWithoutCabalError
+        (Nop, Yes, _)   -> output (False, True, False)
+        (Yes, Idk, _)   -> output (True, False, False)
+        (Yes, Nop, _)   -> output (True, False, False)
+        (Yes, Yes, Idk) -> output (True, True, False)
+        (Yes, Yes, Nop) -> output (True, True, False)
```
vrom911: Heh, I need to think more on this, as I'm not sure yet what is affected by Nix, sorry.

summoner-cli/src/Summoner/Template/Nix.hs
```diff
@@ -0,0 +1,142 @@
...
+import NeatInterpolation (text)
+
+import Summoner.Default (defaultGHC)
+import Summoner.GhcVer (GhcVer (..))
+import Summoner.Settings (NixPkgSet (..), Settings (..), defaultNixPkgSet)
+import Summoner.Tree (TreeFs (..))
+
+import qualified Data.Text as T
+
+nixCompiler :: GhcVer -> Text
```
vrom911: I guess it can be moved to `GhcVer` module. (probably we can make it without explicit pattern matching, but this is not urgent)

summoner-cli/src/Summoner/Template/Nix.hs
```diff
@@ -0,0 +1,142 @@
...
+            , "  url = \"https://github.com/${owner}/${repo}/archive/${rev}.tar.gz\";", endLine
+            , "  inherit sha256;", endLine
+            , "}"
+            , endLine
+            ]
+
+        NixPkgSet { npsOwner, npsRepo, npsRev, npsSha } = case settingsNixPkgSet of { Just n -> n; Nothing -> defaultNixPkgSet }
+
+        nixpkgsNixT :: Text
+        nixpkgsNixT = T.concat
```
vrom911: Why not use `neat-interpolation` here to make it human-readable?

Generally, we sometimes avoid it because of the empty line it can leave depending on we have the empty text or not, but here I don't see that anything like this is present. Probably you had other reasons?

summoner-cli/src/Summoner/Template/Nix.hs
```diff
@@ -0,0 +1,142 @@
...
+            |]
+
+        shellNixT :: Text
+        shellNixT = "(import ./default.nix {}).env" <> endLine
+
+        releaseNixT :: Text
+        releaseNixT = T.concat
+            [ "{"
+            , endLine
+            , T.intercalate endLine
```
vrom911: probably it's better to move int separate function, which will build this matrix?

vrom911:
I don't really get all the question and comment you wrote, so I feel like mostly all that additional actions are not really Summoner-related :thinking: 

About validation of the complex stuff, I'm not sure that we should do this. If ask me I wouldn't let to enter such complex info via terminal. I would just leave the nix support only if there is info in the config, otherwise, it's unmaintainable.. Another issue, that it's not possible to reflect it in TUI. Ideally, the best decision is to have a simple question like "Add Nix? y/n" only if the config file contains all of the required info.

> i think, for the clean source thing, i'd prefer to just use https://github.com/siers/nix-gitignore. it requires that the project have a .gitignore file (it uses the .gitignore to inform how it cleans the source). or there could be a nixCleanSource = ["ignoreThis", "ignoreThis"]. unsure. i think that should be a separate PR from this, as its not integral.

I don't understand, `gitignore` doesn't work for nix? How is this related? `gitignore` file is the one for `GitHub`, I don't really see what is the issue, and why Summoner should solve it?

chessai:
> I don't understand, gitignore doesn't work for nix? How is this related? gitignore file is the one for GitHub , I don't really see what is the issue, and why Summoner should solve it?

Nix has a feature that i will refer to as cleanSource. If you are building a project with nix, since nix is language-agnostic, nix doesn't understand which file types are required as build inputs to your project. That means any changes to thr filesystem within your project will trigger a rebuild if you try to run something like `nix build`, even if all that happened is you added a README or something. Nix's cleanSource takes a bunch of file descriptions, similar to that of .gitignore, and will ignore those for the purpose of buildInputs. So, adding a README or opening a file in vim and having its swap file opened will not trigger a rebuild, only change to actually source or cabal/stack should.

The project i linked generates cleanSource information from a .gitignore, since there's often a huge amount of overlap between the contents of a .gitignore and which files you want ignored by nix.

As for why i thought this should be configured by summoner, it's not hard to configure, could probably just copy over the existing gitignore defaults, and it's something that has to be manually configured every time you set up a nix project nowadays. In short, it's easy to implement and very useful to end users.

chessai:
The generated nix configuration calls `cabal2nix` ('translates' .cabal to a .nix) automatically for the user, so it needs a cabal file for the project, otherwise it can't function.

chessai:
Will respond to other comments shortly, i am on mobile right now and responding to comments is hard.

summoner-cli/src/Summoner/Project.hs
chessai: Don't worry about questions you ask. There are two ways you can use nix w.r.t. dependencies; using an unpinned (system-level) package set, that comes with an installation of nix/nixos, or a pinned package set. the advantage to a pinned package set is that builds become deterministic across multiple machines, since there can be no guarantee that the package set i use is the same as the one you use. The way a nix package set is usually acquired is through github. There are four fields to a nix package set:

* Owner (GitHub owner name)
* Repo (GitHub repo name)
* Revision (GitHub SHA1 revision)
* SHA256 Hash of the package set

for owner $OWNER, repo $REPO, revision $REVISION, and SHA256 hash $SHA, the url for the package set would be at https://github.com/$OWNER/$REPO/archive/$REVISION.tar.gz

You may be wondering where the SHA comes into play. When you attempt to use a package set, you must provide a SHA256 hash of the package set's unpacked NAR archive (more on NAR in a bit). Nix checks that the SHA256 you provide is equivalent to the SHA256 of the package set, and if it is not equivalent, rejects the package set and fails to build. This ensures that only trusted package sets are used. SHA1 is not suitable for this, as has been proven multiple times by the NPM/Golang community. I think a recent feature of stack is to also allow for the sha256 hash of github dependencies, and iirc cabal will soon follow suit, if it hasn't already.

A NAR is a Nix ARchive file. NAR is a (i think strict) subset of tar. TAR, and other archivers, are not deterministic. They add variable-length padding, do not sort files, add timestamps, and a bunch of other stuff. Thus a SHA256 hash of a TAR archive is insufficient. See section 5.2 of eelco's [thesis](https://nixos.org/~eelco/pubs/phd-thesis.pdf).

For the purposes of summoner, I provide a default as `Summoner.Settings.defaultNixPkgSet`, which uses a specific git revision and hash.

You may be wondering how users compute the hash? I touched on this rather vaguely in a comment yesterday. There are two ways.

1) Insert a fake hash with the correct length and nix will complain that it received the wrong hash, and halt the build. It will provide the correct hash. This method is slow and annoying, since it requires you to trigger a build. I don't recommend it normally.
2) use `nix-prefetch-url`, an executable that can give you a bunch of information about urls w.r.t. how they can be used by nix. the most common use case (and the one we might care about) is:

```sh
nix-prefetch-url --unpack https://github.com/$OWNER/$REPO/archive/$REVISION.tar.gz
```

This will take the given nix package set found at that github archive and return its SHA256 hash. The --unpack flag tells it to give you the hash of the unpacked NAR archive. To get the hash of the TAR archive, remove that flag. The comment I made yesterday was saying that to make it easier on users, a potential enhancement is that they only need to provide a revision, and we will compute the hash for them. I have a simple haskell tool that already does this [here](https://github.com/chessai/grab). This feature is not necessary or integral, just an enhancement.

summoner-cli/src/Summoner/Project.hs
chessai: > Thanks for taking your time to answer all these questions! I appreciate this!
> 
> > The generated nix configuration calls cabal2nix ('translates' .cabal to a .nix) automatically for the user, so it needs a cabal file for the project, otherwise, it can't function.
> 
> Stack also needs `cabal` file, and we still can choose one: if you choose only Stack you still have `cabal` file, but also `stack.yaml` files, and Travis config will be only Stack related (without Cabal runs). So the same logic can be applied to Nix & Cabal

Ah, okay, that will probably simplify the 27-case function then.

README.md
```diff
@@ -332,6 +333,7 @@ Here is the list of the options that can be configured to suit your needs. If op
 | `stylish.*`      | Text    | `stylish.file` to provide the absolute file path OR `stylish.url` to download the `.stylish-haskell.yaml` file to use in the project.                                |
 | `contributing.*` | Text    | `contributing.file` to provide the absolute file path OR `contributing.url` download OR `contribuint.link` to link the `CONTRIBUTING.md` file to use in the project. |
 |`[prelude]`       |         |                                                                                                                                                                      |
+|`[nixpkgs]`       |         |                                                                                                                                                                      |
```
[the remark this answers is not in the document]
chessai: ah okay, the formatting did indeed confuse me. understood

chessai:
I think it might be easier for users (and maintainers) if we just use the default if there's no nixpkgs configuration found.

chessai:
need to fix summoner-tui and test suites - tired. sleep now

summoner-cli/src/Summoner/Default.hs
```diff
@@ -41,3 +42,9 @@ currentYear = do
     now <- getCurrentTime
     let (year, _, _) = toGregorian $ utctDay now
     pure $ show year
+
+-- | Default compiler for nix to use.
+defaultNixCompiler :: Text
+defaultNixCompiler = nixCompiler defaultGHC
+
```
qoelet: Remove empty lines?

summoner-cli/src/Summoner/Settings.hs
```diff
@@ -22,6 +25,72 @@ customPreludeT = CustomPrelude
...
+--
+-- https://github.com/NixOS/nixpkgs/archive/fakeRev.tar.gz
+--
+-- where nix would make sure you have the correct SHA256 hash.
+--
+data NixPkgSet = NixPkgSet
+    { npsOwner :: Text
+    , npsRepo  :: Text
+    , npsRev   :: Text
+    , npsSha   :: Text
```
qoelet:
```suggestion
    , npsSha256   :: Text
```

qoelet: Might be nitpicking but I vote for a more explicit naming here

chessai:
do `fullProject` and `smallProject` (in test/golden) need to be manually updated for the tests to pass?

ChShersh:
@chessai 

> do fullProject and smallProject (in test/golden) need to be manually updated for the tests to pass?

Yes, I usually do this. Preferably with some editor that doesn't remove trailing spaces, because `neat-interpolation` sometimes inserts them (see this issue for more details: https://github.com/nikita-volkov/neat-interpolation/issues/23). Current error message from tests is not that easy to decode (because it doesn't tell you much), thus one should be careful when patching tests (but only for now).

chessai:
Unsure why the test suite is failing. It complains about the presence of release.nix, although release.nix should be expected:

```sh
$ cabal new-build
...
Configuring test suite 'summoner-test' for summoner-1.2.0..
Preprocessing test suite 'summoner-test' for summoner-1.2.0..
Building test suite 'summoner-test' for summoner-1.2.0..
[1 of 6] Compiling Prelude
[2 of 6] Compiling Test.DecisionSpec
[3 of 6] Compiling Test.Golden
[4 of 6] Compiling Test.QuestionSpec
[5 of 6] Compiling Test.TomlSpec
[6 of 6] Compiling Main
Linking /home/chessai/Development/summoner/dist-newstyle/build/x86_64-linux/ghc-8.6.3/summoner-1.2.0/t/summoner-test/build/summoner-test/summoner-test ...
Running 1 test suites...
Test suite summoner-test: RUNNING...

mkDefaultYesNoPrompt works correctly
  makes cabal prompt
golden tests
  correctly creates full project FAILED [1]
  correctly creates small project

Failures:

  test/Test/Golden.hs:37:23: 
  1) golden tests correctly creates full project
       expected: []
        but got: ["test/golden/fullProject/release.nix"]

  To rerun use: --match "/golden tests/correctly creates full project/"

Randomized with seed 1737432394

Finished in 0.0030 seconds
3 examples, 1 failure
Test suite summoner-test: FAIL
Test suite logged to:
/home/chessai/Development/summoner/dist-newstyle/build/x86_64-linux/ghc-8.6.3/summoner-1.2.0/t/summoner-test/test/summoner-1.2.0-summoner-test.log
0 of 1 test suites (0 of 1 test cases) passed.
cabal: Tests failed for test:summoner-test from summoner-1.2.0.

```

chessai:
When I remove releaseNixT from the tree, the tests pass. i don't understand why.

summoner-cli/src/Summoner/Template/Nix.hs
```diff
@@ -0,0 +1,131 @@
...
+ 
+        fetchNixpkgsNixT :: Text
+        fetchNixpkgsNixT = T.unlines
+            [ "{ owner  # The owner of the nixpkgs"
+            , ", repo   # The repo of the nixpkgs"
+            , ", rev    # The Git revision of nixpkgs to fetch"
+            , ", sha256 # The SHA256 hash of the unpacked archive"
+            , "}:"
+            , "builtins.fetchTarball {"
+            , "  url = \"https://github.com/${owner}/${repo}/archive/${rev}.tar.gz\";"
```
vrom911: Is this correct? This is just text, so these fields will not be interpolated. If you'd like to use this as variables you should use `text` quasi-quoter here.

summoner-cli/src/Summoner/Template/Nix.hs
```diff
@@ -0,0 +1,131 @@
...
+        nixpkgsNixT :: Text
+        nixpkgsNixT = T.unlines
+            [ ("{ compiler ? \"" <> defaultNixCompiler <> "\" }:")
+            , "with rec {"
+            , "  fetchNixpkgs = import ./fetchNixpkgs.nix;"
+            , "  nixpkgs = fetchNixpkgs {"
+            , ("    owner  = \"" <> npsOwner <> "\";")
+            , ("    repo   = \"" <> npsRepo <> "\";")
+            , ("    rev    = \"" <> npsRev <> "\";")
+            , ("    sha256 = \"" <> npsSha256 <> "\";")
```
vrom911: you can use quasi-quoter here as well. To escape `$` for `$compiler` you can just write `$${compiler}`

vrom911:
@chessai Do you have any update or plans for this PR?

vrom911:
@chessai could you make me the collaborator of your fork so I can fix tests, fix build, resolve conflicts and fix the comments that I can fix? If you would not like to do that I can understand, but in this case, I think the best I can do is to make the PR based on your branch. What do you think?

chessai:
@vrom911 you should have just received an invite link. sorry for the delay. the last thing to be done was resolve the test suite issues.

ChShersh:
> When I remove releaseNixT from the tree, the tests pass. i don't understand why.

@chessai Recently PR with usage of `tree-diff` for displaying better diffs in tests was merged. This should help to find the reason for test failure, if you rebase on `master`.

chessai:
@ChShersh thanks, ok

chessai:
tests pass now. `tree-diff` was extremely helpful in pinpointing the problem, which turned out to be very tiny.

chessai:
Travis is failing because of the TUI

vrom911:
@chessai yes, TUI should fail to build as the `Settings` type has changed, but CLI and TUI sharing this type to support whole functional of the Summoner.

chessai:
Yeah, just glad I know why the tests fail now. I'll get to the TUI soon
hopefully.

vrom911:
Good point @ChShersh. I also noticed some changes recently that as I can understand means that we need the maintainer for changing that particular information when something changes somewhere (which I'm completely unaware about). So I also can't imagine how you or I could be able to keep it up to date or at least fix it when it would break.

chessai:
Note: I am writing this on mobile.

Addressing concerns about maintainability:

Ah, sorry about that. I've been very time-strained lately. Once I got the
CLI working it has been hard to justify allotting time to fixing the TUI,
though now that the tests finally pass (due to help from tree-diff) this
seems more doable.

Please don't hold up progress on summoner for the sake of this PR. A common
concern you both express is that of maintainability: perhaps I could be
added as a GitHub maintainer, primarily for making sure the nix stuff stays
up-to-date. Though this introduces the maintainability concern of only
having one person able to maintain the nix parts of summoner.

Addressing issues raised about current code:

Installing via nix involves two possibilities: from nixpkgs and from source.

From nixpkgs:
nix-env -iA haskellPackages.summoner
(Also works from nix-shell)

From source:
git clone git@github.com:kowainik/summoner && cd summoner
#build & install summoner-cli
cd summoner-cli
nix build && nix-env -i $(readlink -f result)
#build & install summomer-tui
nix build && nix-env -i $(readlink -f result)

tomland is specified in overrides.nix, where we are to state overrides for
our package set. For example, if a package is broken/doesn't build/needs
some unreleased feature, we can just override it. Anything not overridden
is just the default. tomland's test suite did not build (or pass? Can't
remember right now), and the default tomland in the package set was an
older version, while we require tomland >= 0.1. These are the reasons why
tomland is singled out.

Yes, that should be ghc864 now.

Okay, I agree the changelog should just say "add nix scaffolding",
basically.

That cachix script ('cachix.sh') is used to push summoner and its
dependencies to cachix (htpps://cachix.org). Cachix is a very user-friendly
way of setting up public/private caches using nix's unique caching
features. It currently pushes to my company's public cachix, which it
should not do - there should be a 'kowainik' cachix or something. This
script doesn't really add any maintability overhead - it's guaranteed to
work if summoner succeeds to build. So, it can just be part of CI once set
up.

As far as strictness in those fields, I think i need to pull in master's
recent changes.

Additionally, I'm really sorry, I didn't know you were being held up by
this. I'll work on this today. Apologies.

ChShersh:
> Additionally, I'm really sorry, I didn't know you were being held up by
this. I'll work on this today. Apologies.

Ah, no, let me clarify this. It's not urgent to make this PR, we are not planning to add `nix` support into the next release no matter what. So, please, don't work on this today 🙂 What I was trying to say is that I was a little bit uncomfortable about making changes to the `summoner` knowing that they will introduce breaking changes with this PR. However, knowing that you're okay with having breaking changes, we can continue our normal work with `summoner` without worrying 🙂 

As you pointed out, it might not be a good idea to have a single person who knows how to maintain specific part of the application and when other maintainers don't know how to deal with this part. Also, I don't want to put burden of maintaining `summoner` on you knowing how little time you have. Because of that, having `nix` feature merged into `master` will slow down `summoner` developers unless all maintainers are comfortable with all parts of this package. Sometimes it's better to not have some feature implemented rather than having it implemented but not maintainable on the same level as other features. I didn't have time to learn `nix`. I'm using both `cabal` and `stack` for more than 4 years, so, I'm comfortable with maintaining these build tools and I'm aware of possible pitfalls. However, I can't say the same about `nix`. This PR serves as an excellent point of reference on how `nix` can be added to `summoner`. However, this is an extremely major change and should be done carefully.

chessai:
Ah, okay, makes sense. I will pace myself then. Thanks.

PR #293, opened 2019-02-27
Better golden testing using 'tree-diff'

ChShersh:
This PR implements better output of diff in tests when there's mismatch between golden examples and scaffolded by `summoner` project. It uses `tree-diff` library (thanks @phadej for implementing it!).

If there are errors in `fullProject` the output is big. But it's still possible and relatively convenient to find error so it can be fixed easily.


ChShersh:
@vrom911 I force-pushed. I hope everything works now :crossed_fingers:

PR #294, opened 2019-03-05
[#269] Use 'ghc-options' instead of 'warnings'

### LINKED ISSUE #269: Rename `warnings` field in the configuration to `ghc-options`
Currently we have `warnings` field in `.toml` configuration file. But the purpose of this field is to add not only warnings, but also any other GHC flags and options. Thus it makes more sense to name this field as `ghc-options` rather then `warnings`.

Since this is a very breaking change, I propose the following strategy to deal with this:

**Next release**

1. Add `ghc-options` field in addition to `warnings`.
2. Tell message about deprecation of `warnings` field and suggestions to use `ghc-options` instead.

**After release**

1. Drop `warnings` field.

ChShersh:
Resolves #269

## Checklist:

* Update CHANGELOG
    - [x] [summoner-tui/CHANGELOG](https://github.com/kowainik/summoner/blob/master/summoner-tui/CHANGELOG.md).


summoner-cli/CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ The changelog is available [on GitHub][2].
   Allow users to extend the .gitignore file.
 * [#285](https://github.com/kowainik/summoner/issues/285):
   Implement `summon script` command.
+* [#269](https://github.com/kowainik/summoner/issues/269):
+  Introduce `ghc-options` configuration parameter. Deprecate `warnings` TOML parameter.
```
vrom911: Maybe add the clear sentence  like "call for action" — please, rename the field if you use one, it will be deprecated in the very next release

PR #299, opened 2019-04-05
[#280] Support appveyor for cabal

### LINKED ISSUE #280: Appveyor for Cabal?
I see that Cabal has this:
 https://github.com/haskell/cabal/blob/master/appveyor.yml

ChShersh:
Resolves #280

Seems working. See test repo here:

* https://github.com/ChShersh/summon-appvey-test

### DIFF
--- a/summoner-cli/src/Summoner/Template/GitHub.hs
+++ b/summoner-cli/src/Summoner/Template/GitHub.hs
@@ -230,10 +230,47 @@ gitHubFiles Settings{..} =
           - stack test --system-ghc
         |]
 
-
-    -- create appveyor.yml template
     appVeyorYml :: Text
     appVeyorYml =
+        if settingsCabal
+        then appVeyorYmlCabal
+        else appVeyorYmlStack
+
+    appVeyorYmlCabal :: Text
+    appVeyorYmlCabal =
+        [text|
+        install:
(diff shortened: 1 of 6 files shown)


summoner-cli/src/Summoner/Template/GitHub.hs
```diff
@@ -230,10 +230,47 @@ gitHubFiles Settings{..} =
           - stack test --system-ghc
         |]
 
-
-    -- create appveyor.yml template
     appVeyorYml :: Text
     appVeyorYml =
+        if settingsCabal
+        then appVeyorYmlCabal
+        else appVeyorYmlStack
```
vrom911: Probably later we would figure out how to use both, but this is out of the scope of this issue.

ChShersh: I agree. Also, according to my experience, `cabal` in AppVeyor runs faster with these two settings compared. And it works without any problems on the package that we generate. So I prefer to keep `cabal` the default unless you explicitly tell that you don't want to support `cabal`.

Probably `stack` also can work faster, but I'm not using AppVeyor much and don't have much experience with Windows package managers, so can't improve things a lot on my own here. Maybe later somebody will write `Dead simple AppVeyor settings for both Cabal and Stack` 🙂

PR #311, opened 2019-04-11
[WIP] Automatically fill prelude module name

### LINKED ISSUE #255: Automatically fill prelude module name when prelude name is typed
Like in CLI, assume that if the prelude name is `somePrelude` the prelude module name could be `SomePrelude`.

Naman-sopho:
Resolves #255 

Help needed. See this [comment](https://github.com/kowainik/summoner/issues/255#issuecomment-473678107) for details.

## Checklist:

* Update CHANGELOG
- [ ] Keep code style used in the changed files (see [style-guide](https://github.com/kowainik/org/blob/master/style-guide.md#haskell-style-guide) for more details).
- [ ] Use the [`stylish-haskell` file](https://github.com/kowainik/summoner/blob/master/.stylish-haskell.yaml).

### DIFF
--- a/summoner-tui/src/Summoner/Tui.hs
+++ b/summoner-tui/src/Summoner/Tui.hs
@@ -120,14 +120,21 @@ appNew dirs = App
                 withForm ev s (validateForm . ctrlD)
 
             -- Handle active/inactive checkboxes
-            VtyEvent (V.EvKey (V.KChar ' ') []) -> case getCurrentFocus s of
-                Nothing    -> withFormDef ev s
-                Just field -> handleCheckboxActivation ev s field
+            VtyEvent (V.EvKey (V.KChar ' ') []) ->
+              let field = getCurrentFocus s in
+              case field of
+                  Nothing    -> withFormDef ev s
+                  Just prelude@(CustomPreludeName) -> autoFillPreludeModuleName ev s prelude
+                  Just fields -> handleCheckboxActivation ev s fields
             MouseDown n _ _ _ -> handleCheckboxActivation ev s n
 
             -- Handle skip of deactivated checkboxes
             VtyEvent (V.EvKey (V.KChar '\t') []) -> loopWhileInactive ev s
             VtyEvent (V.EvKey V.KBackTab     []) -> loopWhileInactive ev s
+
+            {-VtyEvent (V.EvKey (V.KChar ' ') []) -> case getCurrentFocus s of
+                Nothing    -> withFormDef ev s
+                Just field -> autoFillPreludeModuleName ev s field-}
 
             -- Default action
             _ -> withFormDef ev s
@@ -174,6 +181,17 @@ appNew dirs = App
             Just field -> if not $ isActive (formState newForm) field
                 then loopWhileInactive ev newForm
                 else continue newForm
+
+    -- Handles auto-fill of Prelude module name
+    autoFillPreludeModuleName
+        :: BrickEvent SummonForm e
+        -> KitForm e
+        -> SummonForm
+        -> EventM SummonForm (Next (KitForm e))
+    autoFillPreludeModuleName ev form preludeField = withForm ev (getUpdatedForm form preludeField) mkNewForm
+
+    getUpdatedForm :: KitForm e -> SummonForm -> KitForm e
+    getUpdatedForm form field = form
 
 -- | Draws the form for @new@ command.
 drawNew :: [FilePath] -> KitForm e -> [Widget SummonForm]


summoner-tui/src/Summoner/Tui.hs
```diff
@@ -175,6 +182,17 @@ appNew dirs = App
...
+    -- Handles auto-fill of Prelude module name
+    autoFillPreludeModuleName
+        :: BrickEvent SummonForm e
+        -> KitForm e
+        -> SummonForm
+        -> EventM SummonForm (Next (KitForm e))
+    autoFillPreludeModuleName ev form preludeField = withForm ev (getUpdatedForm form preludeField) mkNewForm
+
+    getUpdatedForm :: KitForm e -> SummonForm -> KitForm e
+    getUpdatedForm form field = form
```
vrom911: You can take as an example this function:
https://github.com/kowainik/summoner/blob/de3b4cbfd85abd77ff86d0746ac81d3b6773651a/summoner-tui/src/Summoner/Tui.hs#L142-L143

You will just need to update another field instead.
Let me know if this helps.

summoner-tui/src/Summoner/Tui.hs
```diff
@@ -120,15 +120,22 @@ appNew dirs = App
                 withForm ev s (validateForm . ctrlD)
 
             -- Handle active/inactive checkboxes
-            VtyEvent (V.EvKey (V.KChar ' ') []) -> case getCurrentFocus s of
-                Nothing    -> withFormDef ev s
-                Just field -> handleCheckboxActivation ev s field
+            VtyEvent (V.EvKey (V.KChar ' ') []) ->
```
ChShersh: I don't think this even should change. Packages in Haskell can't contain whitespaces, so we can just not to do anything at all with the prelude module in this case

summoner-tui/src/Summoner/Tui.hs
Naman-sopho: So, to update a field in the form you need to construct a whole new `SummonKit`, right?
In order to get the updated form, I changed `getUpdatedForm` to use a new `SummonKit` with all fields same except the `preludeModule`.
But I am getting an error
 ```
Variable not in scope:
      gitignore :: Lens.Micro.Type.Getting [Text] (KitForm e1) [Text]
 225 |                                                         , summonKitGitignore    = form ^. gitignore

```

File: summoner-tui/src/Summoner/Tui.hs (hunk 2 above)
Comment (Naman-sopho) [reply summoner-tui/src/Summoner/Tui.hs]: Okay. So we want to trigger this check every time a character in `[a..z] or [A..Z]` is entered.

File: summoner-tui/src/Summoner/Tui.hs (hunk 1 above)
Comment (vrom911) [reply summoner-tui/src/Summoner/Tui.hs]: Oh, no. You don't need to. As you see in the example that I pasted in the previous comment you can use setters to update only fields that you need to update. In your case, you need to update `projectMetaPreludeModule` for which you should use `preludeModule` lens.

Comment (Naman-sopho) [reply summoner-tui/src/Summoner/Tui.hs]: I have pushed an implementation using the `preludeModule` lens to update the field with the value from `preludeName` field. But I am getting the following error on line 201:
```
src/Summoner/Tui.hs:201:16: error:
Couldn't match expected type ‘SummonKit’
                  with actual type ‘f0 SummonKit’
```

This seems like a noobish question, so I tried looking for a solution for quite a few days but could not find any relevant resources with a solution online.

File: summoner-tui/src/Summoner/Tui.hs [hunk 3]
```diff
@@ -175,6 +184,22 @@ appNew dirs = App
...
+        -> EventM SummonForm (Next (KitForm e))
+    autoFillPreludeModuleName ev form preludeField = withForm ev (getUpdatedForm form preludeField) mkNewForm
+
+    getUpdatedForm :: KitForm e -> SummonForm -> KitForm e
+    getUpdatedForm f field =
+      let
+        input = f ^. projectMeta . preludeName
+        form = formState f & preludeModule input
+      in
+        mkForm form
```
vrom911: I didn't checked, but something like this should do the job:

```haskell
import Summoner.Text (packageToModule)

updatePreludeModule :: KitForm e -> SummonForm -> KitForm e
updatePreludeModule = mkForm $ formState f & projectMeta . preludeModule .~ uppercasePrelude
  where
    uppercasePrelude :: Text
    uppercasePrelude = packageToModule $ formState f ^. (preludeMeta . preludeName)
```

summoner-tui/src/Summoner/Tui.hs
chshersh: @Naman-sopho We want to trigger this event on `[a..z], [A..Z], [0..9], -`. This are the characters allowed in the package name. And then the module name field should be automatically filled using the value of prelude package name

vrom911:
Hey @Naman-sopho, I just want to check if my last comments make sense. Let me know if you have any questions.

Naman-sopho:
Yes. I have implemented the update logic correctly. But it is only triggered when the space bar is pressed(as is the current implementation). I am working on the triggering logic currently, need to trigger that with every alphanumeric click.

chshersh:
@Naman-sopho Hi! Any updates on this issue?

vrom911:
@Naman-sopho friendly ping :slightly_smiling_face:

PR #324, opened 2019-05-29
[#323] Allow tree-diff-0.1

### LINKED ISSUE #323: Allow tree-diff-0.1
https://github.com/commercialhaskell/stackage/issues/4593

vrom911:
Resolves #323

Seems like we can just ake a revision at Hackage


vrom911:
Revised on Hackage.

PR #331, opened 2019-07-11
[#328] Upgrade to tomland-1.1.0.0

### LINKED ISSUE #328: Bump up tomland to 1.1.0.0
https://github.com/commercialhaskell/stackage/issues/4703

vrom911:
Resolves #328

### DIFF
--- a/summoner-cli/summoner.cabal
+++ b/summoner-cli/summoner.cabal
@@ -109,7 +109,7 @@ library
                      , shellmet ^>= 0.0.1
                      , text ^>= 1.2.3.0
                      , time ^>= 1.8
-                     , tomland ^>= 1.0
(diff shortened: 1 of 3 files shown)


summoner-cli/summoner.cabal
```diff
@@ -109,7 +109,7 @@ library
                      , shellmet ^>= 0.0.1
                      , text ^>= 1.2.3.0
                      , time ^>= 1.8
-                     , tomland ^>= 1.0
+                     , tomland >= 1.0
```
chshersh: I think having upper bounds is always nice, even for our packages. We can accidentally break something, can happen with everyone, and we don't want our own packages to break...
```suggestion
                     , tomland >= 1.0 && < 1.2
```

Issue #334, opened 2019-08-17
Avoiding symlinks in cabal source dists


srid:
I am working on adding a `default.nix` to summoner so that it can be built using Nix.

But there is a problem. The README/LICENSE of the two cabal packages symlink to their targets _outside_ of their directories (i.e., the parent directory). This causes issues with Nix build, because Nix copies the symlinks as is and `cabal build` would later fail.

I propose a couple of changes:

- Replace symlinks with regular files (toplevel/README.md can be a symlink, for Github rendering)
- Respect DRY by having, say, `summoner-tui/README.md` refer to the README of `summoner-cli/README.md`. This would reflect on the hackage docs correspondingly (currently there is README duplication between the -cli and -tui hackage docs because of the symlink).

What do you think? I can provide more context/ rationale for this change if needed. I cannot think of alternatives that will keep things as simple, but I'm open to suggestions.

srid:
To see how it all would look like, see https://github.com/srid/summoner/tree/nixify

vrom911:
Hey @srid, thanks for working on Nix integration!

Have you tried to address this Nix issue to Nix itself? I do see problems with both of the proposed solutions, which is not making life easier. I would prefer to keep things as it is and to fix the problem with nix itself if it's possible. Please share your updates on the topic, thanks! :slightly_smiling_face:

srid:
It is not really an issue with Nix per se, as having symlinks pointing to files outside the source directory is considered impure.

I could customize the `default.nix` to manually copy over the target files into the source derivation; I'll give that a try.

srid:
Worked around in https://github.com/kowainik/summoner/pull/335/commits/b57710a3573dc9b3021797aaa2e48ef08e49a88f

PR #335, opened 2019-08-19
Add support for building summoner using Nix

### LINKED ISSUE #150: Add `.nix` file to `summoner` as well

srid:
Resolves #150

I still need to fix an issue with README/LICENSE symlink as indicated #334. Once that's done this PR will be ready to be merged. But first, I'd like know if this PR is welcome.

## Checklist:

* Update CHANGELOG

### DIFF
--- a/README.md
+++ b/README.md
@@ -583,7 +583,14 @@ for Cabal and
 stack build
 ```
 
-for Stack.
+for Stack and
+
+```shell=
+nix-build -A summoner     # To build the CLI only
+nix-build -A summoner-tui # To build the TUI
+```
+
+for Nix.
 
 ### Test [↑](#structure)
 
@@ -595,10 +602,16 @@ To actually run tests you need to run:
 cabal new-test all
 ```
 
-or
+or, if using Stack,
 
 ```shell
 stack test
+```
+
+or, if using Nix,
+
+```shell
+nix-shell --run 'cabal new-test all'
 ```
 
 ### Run [↑](#structure)
@@ -610,11 +623,18 @@ cabal new-exec summon -- SOME_COMMAND
 cabal new-exec summon-tui -- SOME_COMMAND
 ```
 
-or
+or, if using Stack,
 
 ```shell
 stack exec summon -- SOME_COMMAND
 stack exec summon-tui -- SOME_COMMAND
+```
+
+of, if using Nix,
+
+```shell
+nix-shell --run 'cabal new-exec summon -- SOME_COMMAND'
+nix-shell --run 'cabal new-exec summon-tui -- SOME_COMMAND'
 ```
 
 ## Changelog [↑](#structure)
--- a/README.md
+++ b/README.md
@@ -630,7 +630,7 @@ stack exec summon -- SOME_COMMAND
 stack exec summon-tui -- SOME_COMMAND
 ```
 
-of, if using Nix,
+or, if using Nix,
 
 ```shell
 nix-shell --run 'cabal new-exec summon -- SOME_COMMAND'
--- a/README.md
+++ b/README.md
@@ -633,8 +633,8 @@ stack exec summon-tui -- SOME_COMMAND
 or, if using Nix,
 
 ```shell
-nix-shell --run 'cabal new-exec summon -- SOME_COMMAND'
-nix-shell --run 'cabal new-exec summon-tui -- SOME_COMMAND'
+nix-shell --run 'cabal new-run summon -- SOME_COMMAND'
+nix-shell --run 'cabal new-run summon-tui -- SOME_COMMAND'
 ```
 
 ## Changelog [↑](#structure)
--- a/default.nix
+++ b/default.nix
@@ -21,12 +21,8 @@ let
     summoner = ./summoner-cli;
     summoner-tui = ./summoner-tui;
 
-    relude = fetchGitHubArchive "kowainik" "relude"
-      "55968311244690f5cc8b4484a37a63d988ea2ec4";
     tomland = fetchGitHubArchive "kowainik" "tomland"
-      "8ef78e5fcfd7055b1db6402713d27a0aa42a82f8";
-    shellmet = fetchGitHubArchive "kowainik" "shellmet"
-      "36149eb0eb2b81916a93cdb92f3cb949d2eb9d23";
+     "8ef78e5fcfd7055b1db6402713d27a0aa42a82f8";
     optparse-applicative = fetchGitHubArchive "pcapriotti" "optparse-applicative"
       "5478fc16cbd3384c19e17348a17991896c724a3c";
     parser-combinators = fetchGitHubArchive "mrkkrp" "parser-combinators"
--- a/default.nix
+++ b/default.nix
@@ -2,7 +2,7 @@
   # the time of this writing has GHC 8.6.5 as the default compiler (matching the
   # one used by stack.yaml). Use https://howoldis.herokuapp.com/ to determine
   # the current rev.
-  pkgs ? import (builtins.fetchTarball "https://github.com/nixos/nixpkgs/archive/002b853782e.tar.gz") {}
+  pkgs ? import (builtins.fetchTarball "https://github.com/nixos/nixpkgs/archive/ed07403382c.tar.gz") {}
   # Which GHC compiler to use.
   # To determine the list of compilers available run:
   #   nix-env -f "<nixpkgs>" -qaP -A haskell.compiler
@@ -40,8 +40,6 @@ let
       "36149eb0eb2b81916a93cdb92f3cb949d2eb9d23";
     optparse-applicative = fetchGitHubArchive "pcapriotti" "optparse-applicative"
       "5478fc16cbd3384c19e17348a17991896c724a3c";
-    parser-combinators = fetchGitHubArchive "mrkkrp" "parser-combinators"
-      "4262c00ef70cc30bfc56db0a10c37d88ad88fe1a";
   });
 
   # Summoner project shell.
--- a/default.nix
+++ b/default.nix
@@ -46,7 +46,6 @@ let
 
   # Summoner project shell.
   projectShell = projectDrv.shellFor {
-    withHoogle = true;
     packages = p:
       [ p.summoner
         p.summoner-tui
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,50 @@
+{ # The git revision here corresponds to the nixpkgs-unstable channel, which at
+  # the time of this writing has GHC 8.6.5 as the default compiler (matching the
+  # one used by stack.yaml). Use https://howoldis.herokuapp.com/ to determine
+  # the current rev.
+  pkgs ? import (builtins.fetchTarball "https://github.com/nixos/nixpkgs/archive/002b853782e.tar.gz") {}
+  # Which GHC compiler to use.
+  # To determine the list of compilers available run:
+  #   nix-env -f "<nixpkgs>" -qaP -A haskell.compiler
+, compiler ? "default"
+}:
+let
+  haskellPackages =
+    if compiler == "default"
+      then pkgs.haskellPackages
+      else pkgs.haskell.packages.${compiler};
+  fetchGitHubArchive = owner: repo: rev:
+    builtins.fetchTarball "https://github.com/${owner}/${repo}/archive/${rev}.tar.gz";
+
+  # Summoner project derivation.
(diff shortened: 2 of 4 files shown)


default.nix
```diff
@@ -0,0 +1,50 @@
...
+    relude = fetchGitHubArchive "kowainik" "relude"
+      "55968311244690f5cc8b4484a37a63d988ea2ec4";
+    tomland = fetchGitHubArchive "kowainik" "tomland"
+      "8ef78e5fcfd7055b1db6402713d27a0aa42a82f8";
+    shellmet = fetchGitHubArchive "kowainik" "shellmet"
+      "36149eb0eb2b81916a93cdb92f3cb949d2eb9d23";
+    optparse-applicative = fetchGitHubArchive "pcapriotti" "optparse-applicative"
+      "5478fc16cbd3384c19e17348a17991896c724a3c";
+    parser-combinators = fetchGitHubArchive "mrkkrp" "parser-combinators"
+      "4262c00ef70cc30bfc56db0a10c37d88ad88fe1a";
```
vrom911: I wonder if we would use `LTS-14.1` where all of these packages are present, we won't need these lines anymore? (Sorry, I'm not using Nix, so I don't know how it works...)

default.nix
```diff
@@ -0,0 +1,50 @@
...
+  projectShell = projectDrv.shellFor {
+    withHoogle = false;
+    packages = p:
+      [ p.summoner
+        p.summoner-tui
+      ];
+    buildInputs = with projectDrv;
+      [ cabal-install
+        ghcid
+        pkgs.gitAndTools.hub
```
vrom911: Is this needed to build the project? If no and these are prerequisites, then for the work of the tool `git` and `curl` are also required.

README.md
```diff
@@ -610,13 +623,20 @@ cabal new-exec summon -- SOME_COMMAND
...
 
-or
+or, if using Stack,
 
 ```shell
 stack exec summon -- SOME_COMMAND
 stack exec summon-tui -- SOME_COMMAND
 ```
 
+of, if using Nix,
```
chshersh: Minor typo
```suggestion
or, if using Nix,
```

README.md
```diff
@@ -610,13 +623,20 @@ cabal new-exec summon -- SOME_COMMAND
...
+
+```shell
+nix-shell --run 'cabal new-exec summon -- SOME_COMMAND'
+nix-shell --run 'cabal new-exec summon-tui -- SOME_COMMAND'
```
chshersh: Am I right that these commands also will install all required tools for `summoner` (like `git`, `hub` and `curl`) and people can run `summoner` just by this command without needing to install anything except `nix`?

default.nix
```diff
@@ -0,0 +1,50 @@
...
+      "36149eb0eb2b81916a93cdb92f3cb949d2eb9d23";
+    optparse-applicative = fetchGitHubArchive "pcapriotti" "optparse-applicative"
+      "5478fc16cbd3384c19e17348a17991896c724a3c";
+    parser-combinators = fetchGitHubArchive "mrkkrp" "parser-combinators"
+      "4262c00ef70cc30bfc56db0a10c37d88ad88fe1a";
+  });
+
+  # Summoner project shell.
+  projectShell = projectDrv.shellFor {
+    withHoogle = false;
```
chshersh: Why does this need to be set explicitly to `false`?

default.nix
chshersh: I also wonder why all these source dependencies are required? If it's possible to update dependencies in `summoner` and simplify `nix` config, I'm all for it. I also would like to know how to determine whether I need such source dependencies or not :slightly_smiling_face:

default.nix
srid: It is not needed for building the binaries. And I'll add `git` and `curl` to the list. 

These are just runtime dependencies; you can use `nix-shell` to run summoner without having to install git/hub/curl globally (one of the advantages of using nix is that you can manage *non-Haskell* packages in your project environment!).

README.md
srid: That's correct. :-)

---

You'll notice that the last line of `default.nix` is:

```
if pkgs.lib.inNixShell then projectShell else projectDrv 
```

If you run `nix-shell` it will use `projectShell` (thereby all the extra dependencies specified in `buildInputs`); but if you run `nix-build` it will only use `projectDrv` (which is basically haskell packages needed to build the summoner binary).

vrom911:
Also, @srid, do you by chance know the way we could check Nix build on CI as well? We do have both Cabal and Stack checks with Travis now. If you know an easy way to add Nix let us know!

default.nix
srid: It was an oversight (was debugging something). It's fixed now--the value is now `true` so that `nix-shell run 'hoogle serve'` gives you a local Hoogle server with documentation on dependencies.

srid:
By the way, thank you for asking questions related to Nix (I'll get to the pending questions as soon as I get sufficient information), as these questions give me ideas as to what more I can add to [my Nix tutorial](https://notes.srid.ca/haskell-nix) to make it comprehensive enough for beginners.

default.nix
srid: @vrom911 Here's what I gathered from chatting with nix folks on fpslack:

> It's a bit of a tangled process. One part of it is all-cabal-hashes, which is occasionally updated from here: https://github.com/commercialhaskell/all-cabal-hashes/releases
Part of it is hackage2nix: https://github.com/NixOS/cabal2nix/tree/master/hackage2nix (edited) 
https://github.com/NixOS/nixpkgs/commit/5ad0677aec2455e664992d93d063d5cb34a32e2f
And then part of it is this: https://github.com/NixOS/nixpkgs/blob/5ad0677aec2455e664992d93d063d5cb34a32e2f/pkgs/development/haskell-modules/configuration-hackage2nix.yaml
Which I suppose is generated from an LTS, probably

(That doesn't look like a satisfactory explanation unfortunately, but it is what it is ...)

The only way to modify the package set is to either fork nixpkgs ourselves, or wait for upstream to update it. In either case, nixpkgs will be pinned to a specific git revision (for reproduciblity).

@chshersh 
> I also wonder why all these source dependencies are required? 

They are required (except relude and shellmet, which I've just removed), because the version in pinned nixpkgs is either older (than what the cabal file asks for) or marked as broken.

>  I also would like to know how to determine whether I need such source dependencies or not 

Nix will complain about missing or broken packages. 

For example, running `nix-build` (without specifying these overrides) produced this:

image

`tomland` is marked as broken; so I went and did an override of it in default.nix.

And elsewhere it produced this:

image

The version of `parser-combinators` in the pinned nixpkgs is too old; so I did an override of it in default.nix by using a version that matches the version bounds asked by the cabal file.

IIUC this is similar to the use of `extra-deps` in Stack.

srid:
> Also, @srid, do you by chance know the way we could check Nix build on CI as well? We do have both Cabal and Stack checks with Travis now. If you know an easy way to add Nix let us know!

Travis [supports Nix](https://docs.travis-ci.com/user/languages/nix/), but you would have to use `language: nix`. Is stack a requirement in current travis workflow? If not, maybe we can exclusively use Nix for building, but then I don't have full context behind your CI setup.

Also does anyone know why the CI build is failing? My PR should not in theory be breaking CI.

srid:
I've resoled the symlink issue. This PR is no longer WIP. 

If the travis part can be addressed as separate PR could you re-review the change @vrom911 @chshersh ?

default.nix
```diff
@@ -0,0 +1,65 @@
...
+      "36149eb0eb2b81916a93cdb92f3cb949d2eb9d23";
+    optparse-applicative = fetchGitHubArchive "pcapriotti" "optparse-applicative"
+      "5478fc16cbd3384c19e17348a17991896c724a3c";
+    parser-combinators = fetchGitHubArchive "mrkkrp" "parser-combinators"
+      "4262c00ef70cc30bfc56db0a10c37d88ad88fe1a";
+  });
+
+  # Summoner project shell.
+  projectShell = projectDrv.shellFor {
+    withHoogle = true;
```
chshersh: I would like to have a minimal working config. This will make it easier to understand and maintain :slightly_smiling_face: So if this option is not required and enabling it can potentially slow down build in addition to making the configuration more verbose, I would prefer to remove this line.

vrom911:
> I can add to my Nix tutorial to make it comprehensive enough for beginners.
Wow, cool stuff! Added to my reading list for learning how to Nix. Thanks for doing this :slightly_smiling_face:

default.nix
vrom911: Interesting. Thanks for writing this up!
I guess we should update the pinned version of `nixpkgs` with every new GHC releases or it better to update it more frequent?

chshersh:
@srid I believe I've figured out the Travis CI support for `nix`. I guess you can add something like this to the existing `.travis.yml` to this PR with `nix` on CI:

https://github.com/chshersh/type-errors-pretty/blob/cf63c3363c6a681deae1843eca8524a74df53b12/.travis.yml#L24-L26

Except it will be `cabal v2-test all` instead of `cabal new-test`.

default.nix
srid: Sure, removed.

default.nix
srid: > I guess we should update the pinned version of nixpkgs with every new GHC releases or it better to update it more frequent?

I think updating for every new GHC release is good. A newer nixpkgs may also have newer versions of Haskell packages.

As long as `compiler == "default"` - the Haskell packages will be downloaded from nix cache instead of needing to build locally.

srid:
Thanks @chshersh ! I've added Nix to .travis.yml ... looks like it is building successfully.

chshersh:
@srid Thanks for your effort so far! The build for Nix failed. But not because of build failure, but because it exceeded the time limit (50 minutes). From the build log, I see that it builds every package from sources including Haddock instead of using binary cache. That's why it takes so long to build.

In one of the comments you wrote:

> As long as compiler == "default" - the Haskell packages will be downloaded from nix cache instead of needing to build locally.
 
I've checked this website, and it shows that both `nixos-unstable` and `nixos-19.03` were updated within the last day:

* https://howoldis.herokuapp.com/

Is it somehow relevant to the fact that every package is built instead of reusing binary cache?
Also, if builds take too long, we might consider using and configuring `cachix` to speed up the CI... Though I do not fully understand how `cachix` works precisely.

srid:
> Is it somehow relevant to the fact that every package is built instead of reusing binary cache?

No. *All* revisions (i.e., even older ones) of nixpkgs are in fact cached. 

The reason why Nix is building so many Haskell packages is because we are overriding `optparse-applicative` (summoner needs 0.15; but that's not in nixpkgs). 

> Also, if builds take too long, we might consider using and configuring cachix to speed up the CI... 

Yea, setting up cache is a very good idea.

> Though I do not fully understand how cachix works precisely.

It's pretty simple to use; I encourage you to give it a try.

In the travis script we only need to prepend the nix-build cli with `cachix use kowainik`, which makes the subsequent nix-build use the specified cache. 

To push to/update the cache say from your dev machine, you would do `nix-build -A summoner-cli | cachix push kowainik`. 

Incidentally I personally use http://hercules-ci.com (instead of Travis) for CI, thereby using my home-machine to run the CI jobs, effectively obviating time limits like that of Travis (it takes >50 mins only the first time, the subsequent runs would use cached packages from local store).

srid:
Actually the CI can also _push_ to the cachix store, which has the benefit of allowing both developers and users to utilize the cache for building summoner!

Here's an example: https://github.com/cachix/cachix/blob/master/.travis.yml#L14

srid:
@vrom911 Sure, I'd love that!

srid:
> are you interested in helping Summoner to improve (with Nix and not only!)?

Guys, I'd now like to work on Summoner to add proper Nix support (both at project and script level). Could you direct me to any existing discussion, issues, specs, etc. on this?

chshersh:
@srid Sure, we appreciate any help a lot! @vrom911 and I already have plans in our roadmap to update details regarding nix support under corresponding issue and write our thoughts there in some near future.

andreas-ebot7:
Is it currently possible to install summoner with `nix`? :)

Looks like it's currently marked as broken in hackage-packages

chshersh:
@andreas-ebot7 I'm not sure since CI for Nix hasn't been implemented. So `default.nix` could become outdated. I'm not using `nix` personally, I'm building with `cabal`, so it works for me.

Issue #327, opened 2019-06-01
Improve Appveyor config


vrom911:
- [x] https://github.com/commercialhaskell/stack/issues/4535
- [x] Add `--enable-test` for Cabal

chshersh:
One more useful example for Cabal:

- [ ] https://github.com/haskell-hvr/windns/blob/master/.appveyor.yml

PR #347, opened 2019-10-09
Refactor AppVeyor CI config

chshersh:
AppVeyor CI fails for some weird reason. This is happening because we're using the latest `cabal` and `cabal-3.0` doesn't work (yet). Instead of figuring out, why it doesn't work, I just fixed `cabal` version to 2.4 and I also used this `appveyor.yml` script from this blog post:

* https://hub.zhox.com/posts/introducing-haskell-dev/

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,24 +1,34 @@
-install:
-  # Using '-y' and 'refreshenv' as a workaround to:
-  # https://github.com/haskell/cabal/issues/3687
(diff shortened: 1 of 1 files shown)


chshersh:
@vrom911 I think no... We have a separate issue for patching generated `appveyor.yml` file: https://github.com/kowainik/summoner/issues/327

We can use this one for `cabal` and for default and copy from `relude` for `stack`-only projects.
Also, let's see how good it is and how well it builds for some time :slightly_smiling_face:

PR #352, opened 2019-10-16
Allow users to add newline character in text fields

### LINKED ISSUE #253: Make `Enter` work inside `Description` field in TUI
Currently `Description` input field contains two lines, but <kbd>Enter</kbd> doesn't work there because we create projects on <kbd>Enter</kbd>. It's also not that clear how to handle this situation...

yigitozkavci:
Resolves #253.

This PR enables users to add a newline to multi-line text fields by pressing <kbd>Alt</kbd>+<kbd>Enter</kbd>.

This is not the most intuitive key combination but `brick` currently cannot guarantee modifiers other than <kbd>Alt</kbd> (see https://github.com/jtdaugherty/brick/issues/32).

I tried upgrading `brick` from `0.47.1` to `0.50` but situation is the same, unfortunately.

## Checklist:

* Update CHANGELOG
    - [X] [summoner-cli/CHANGELOG](https://github.com/kowainik/summoner/blob/master/summoner-cli/CHANGELOG.md).
    - [X] [summoner-tui/CHANGELOG](https://github.com/kowainik/summoner/blob/master/summoner-tui/CHANGELOG.md).
- [X] Keep code style used in the changed files (see [style-guide](https://github.com/kowainik/org/blob/master/style-guide.md#haskell-style-guide) for more details).
- [X] Use the [`stylish-haskell` file](https://github.com/kowainik/summoner/blob/master/.stylish-haskell.yaml).
- [X] Update documentation (README, haddock) if required.
- [X] Create a new test project using `summoner` and check that the changes work as expected.

### DIFF
--- a/summoner-cli/src/Summoner/Project.hs
+++ b/summoner-cli/src/Summoner/Project.hs
@@ -8,6 +8,7 @@ module Summoner.Project
        ) where
 
 import Data.List (intersect)
+import qualified Data.Text as T
 import NeatInterpolation (text)
 import Shellmet ()
 import System.Directory (setCurrentDirectory)
@@ -209,7 +210,7 @@ doGithubCommands Settings{..} = do
     "git" ["add", "."]
     "git" ["commit", "-m", "Create the project"]
     unless settingsNoUpload $ do
-        "hub" $ ["create", "-d", settingsDescription, settingsOwner <> "/" <> settingsRepo]
+        "hub" $ ["create", "-d", (T.intercalate " " . lines) settingsDescription, settingsOwner <> "/" <> settingsRepo]
              ++ ["-p" | settingsPrivate]  -- Create private repository if asked so
          -- Upload repository to GitHub.
         "git" ["push", "-u", "origin", "master"]
--- a/summoner-cli/src/Summoner/Project.hs
+++ b/summoner-cli/src/Summoner/Project.hs
@@ -8,7 +8,6 @@ module Summoner.Project
        ) where
 
 import Data.List (intersect)
-import qualified Data.Text as T
 import NeatInterpolation (text)
 import Shellmet ()
 import System.Directory (setCurrentDirectory)
@@ -210,7 +209,7 @@ doGithubCommands Settings{..} = do
     "git" ["add", "."]
     "git" ["commit", "-m", "Create the project"]
     unless settingsNoUpload $ do
-        "hub" $ ["create", "-d", (T.intercalate " " . lines) settingsDescription, settingsOwner <> "/" <> settingsRepo]
+        "hub" $ ["create", "-d", (unwords . lines) settingsDescription, settingsOwner <> "/" <> settingsRepo]
              ++ ["-p" | settingsPrivate]  -- Create private repository if asked so
          -- Upload repository to GitHub.
         "git" ["push", "-u", "origin", "master"]
--- a/summoner-cli/src/Summoner/Project.hs
+++ b/summoner-cli/src/Summoner/Project.hs
@@ -209,7 +209,7 @@ doGithubCommands Settings{..} = do
     "git" ["add", "."]
     "git" ["commit", "-m", "Create the project"]
     unless settingsNoUpload $ do
-        "hub" $ ["create", "-d", (unwords . lines) settingsDescription, settingsOwner <> "/" <> settingsRepo]
+        "hub" $ ["create", "-d", settingsDescription, settingsOwner <> "/" <> settingsRepo]
              ++ ["-p" | settingsPrivate]  -- Create private repository if asked so
          -- Upload repository to GitHub.
         "git" ["push", "-u", "origin", "master"]
--- a/summoner-cli/src/Summoner/Template/Cabal.hs
+++ b/summoner-cli/src/Summoner/Template/Cabal.hs
@@ -13,7 +13,6 @@ import Summoner.Settings (CustomPrelude (..), Settings (..))
 import Summoner.Text (intercalateMap, packageToModule)
 import Summoner.Tree (TreeFs (..))
 
-import qualified Data.List.NonEmpty as NE
 import qualified Data.Text as T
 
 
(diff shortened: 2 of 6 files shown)


yigitozkavci:
> I've tested it locally and I got this in the cabal file, I don't think that this is correct indentation:
Yes, I didn't realize this. I will split the text on cabal file creation then. Ideally I want this feature to be applicable to every text field. I'll see if it's possible to do without introducing invasive changes.

> Also, I'm not sure if multiline text works with the GitHub project description...
I wasn't using `hub` so I couldn't test it. I'll make sure it works before asking for the next review.

yigitozkavci:
Hi @vrom911,

Thanks for the review.

I believe cabalfile and Github descriptions are now fixed and here is a recording:
https://asciinema.org/a/xxkMVW9QXmiEFnjq1iZrwyxGU

At this point I'm inclined to make `settingsDescription`'s type `[Text]` instead of `Text` to make it obvious that it's a multi-line text but I didn't want to introduce larger diff. What do you think?

vrom911:
Thanks for the demo again :heart: 

> I wasn't using hub so I couldn't test it. I'll make sure it works before asking for the next review.

Sad to hear, that you don't actually use the project... I hope that you could give it a try one day, as you're already a contributor :smile: 


Regarding the changes, I believe that the newline used in TUI is for line-fit only, it shouldn't be translated to cabal file as it doesn't bring any profit, but only could introduce parsing errors.

Another point to that is the fact that we want to always have equal features for TUI and CLI (where possible). So making double-line description (this is the limitation at the moment as I see) for TUI is a bit strange, don't you think? But as you're not a user, let's see what @chshersh  is thinking about this, ping @chshersh

summoner-cli/src/Summoner/Project.hs
```diff
@@ -209,7 +210,7 @@ doGithubCommands Settings{..} = do
     "git" ["add", "."]
     "git" ["commit", "-m", "Create the project"]
     unless settingsNoUpload $ do
-        "hub" $ ["create", "-d", settingsDescription, settingsOwner <> "/" <> settingsRepo]
+        "hub" $ ["create", "-d", (T.intercalate " " . lines) settingsDescription, settingsOwner <> "/" <> settingsRepo]
```
chshersh: I think it's the same as using `unwords`
```suggestion
        "hub" $ ["create", "-d", (unwords . lines) settingsDescription, settingsOwner <> "/" <> settingsRepo]
```

summoner-cli/src/Summoner/Template/Cabal.hs
```diff
@@ -29,14 +30,20 @@ cabalFile Settings{..} = File (toString settingsRepo ++ ".cabal") cabalFileConte
...
     cabalHeader = unlines $
         [ "cabal-version:       " <> defaultCabal
         , "name:                " <> settingsRepo
         , "version:             0.0.0.0"
-        , "synopsis:            " <> settingsDescription
-        , "description:         " <> settingsDescription ] ++
+        , "synopsis:            " <> NE.head settingsDescriptionLines ] ++
+        [ "                     " <> desc | desc <- NE.tail settingsDescriptionLines ] ++
+        [ "description:         " <> NE.head settingsDescriptionLines ] ++
+        [ "                     " <> desc | desc <- NE.tail settingsDescriptionLines ] ++
```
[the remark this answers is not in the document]
yigitozkavci: It makes sense to have it in a single line, which is something we do for Github descriptions as well. Thanks!

summoner-cli/src/Summoner/Project.hs
yigitozkavci: Done, thanks!

yigitozkavci:
> I see that (strangely written) `unword $ lines settingsDescription` is used in three different places. This could be improved

First thing that comes to my mind is to add it to `summoner-tui/src/Prelude` but that would only introduce indirection for the rea
I wouldn't want to go to definition of another function just to see `unword . lines`. If you disagree I can just add it to Prelude though.

chshersh:
@yigitozkavci You can put this function into `Summoner.Text` module. You can see that this module already contains general-purpose functions to work with `Text`.

vrom911:
Yes, I disagree, I think that `settingsDescription` should never have newline inside, and all the manipulations on TUI side should be done before it goes to settings :slightly_smiling_face:

chshersh:
Comment by @vrom911 makes more sense to me as well. After changing the code and removing `NonEmpty` I see more clearly that there's no need to introduce this function in the core package and all manipulations can be done on the TUI side.

yigitozkavci:
> I think that settingsDescription should never have newline inside, and all the manipulations on TUI side should be done before it goes to settings

As @chshersh says this became obvious after having the same processing logic for Cabal and Github files. Now we process this string before creating Settings type.

PTAL again.

summoner-tui/src/Summoner/Tui/Validation.hs
```diff
@@ -48,6 +49,15 @@ handleAutofill f =
           setFormFocus CustomPreludeName $ mkForm new_state
     _ -> f
 
+-- | Adds a newline for text fields.
+newLine :: KitForm e -> KitForm e
+newLine =
+  handleNewLine ProjectDesc (project . desc)
+  where
+    handleNewLine formField fieldLens f =
```
vrom911: I guess there is no need to introduce function here, you can inline it in the `newLine` function :slightly_smiling_face: But I don't mind having it, however, by our style guide the type should be written as well.

yigitozkavci: I've added a type signature now.

I didn't want to inline it because I want this function to be applicable to other fields in the future. This way it can be done just by composing different newline handlers on `KitForm e`.

summoner-tui/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ The changelog is available [on GitHub][2].
   Improve `summon show ghc` output
 * [#255](https://github.com/kowainik/summoner/issues/255)
   Autofill custom prelude module name when package name is being typed.
+* [#253](https://github.com/kowainik/summoner/issues/253):
+  Fix not being able to add newline to multi-line fields.
```
vrom911: I believe it's not the exact wording. Let's say explicitly what it does, like "Fix new line in the Description field of the form", what do you think?

summoner-tui/src/Summoner/Tui/Validation.hs
vrom911: Thanks for thinking about that. But currently, we don't have any other multiline or any big-enough fields for that, and nothing we are planning in the nearest future. Do you have any particular thoughts on that?

summoner-tui/CHANGELOG.md
yigitozkavci: Agreed, done :)

summoner-tui/src/Summoner/Tui/Validation.hs
yigitozkavci: > nothing we are planning in the nearest future. Do you have any particular thoughts on that?

I would still prefer having a generic "new line handler" function but if there is no such plans I'm okay with converting it to handle only description. I'm also changing function's name since it's now specific to project description.

PR #355, opened 2019-10-17
Validate prelude package name in TUI

### LINKED ISSUE #354: Validate prelude package name
As per [cabal spec](https://www.haskell.org/cabal/users-guide/developing-packages.html#package-names-and-versions) package names are subject to a well-defined format. We should add a validation for it in summoner-tui.

This is an issue created as an action item from #353.

Also, we could check the module name.

Parts need to be covered:

- [ ] CLI
- [x] Config
- [ ] TUI

yigitozkavci:
Part of #354 

Demo: https://asciinema.org/a/G5uCtFcVIzfNdUeSX7TvHdGe4

## Checklist:

* Update CHANGELOG
    - [X] [summoner-cli/CHANGELOG](https://github.com/kowainik/summoner/blob/master/summoner-cli/CHANGELOG.md).
    - [X] [summoner-tui/CHANGELOG](https://github.com/kowainik/summoner/blob/master/summoner-tui/CHANGELOG.md).
- [X] Keep code style used in the changed files (see [style-guide](https://github.com/kowainik/org/blob/master/style-guide.md#haskell-style-guide) for more details).
- [X] Use the [`stylish-haskell` file](https://github.com/kowainik/summoner/blob/master/.stylish-haskell.yaml).
- [X] Update documentation (README, haddock) if required.
- [X] Create a new test project using `summoner` and check that the changes work as expected.

### DIFF
--- a/summoner-tui/src/Summoner/Tui/Validation.hs
+++ b/summoner-tui/src/Summoner/Tui/Validation.hs
@@ -10,14 +10,13 @@ module Summoner.Tui.Validation
        ) where
 
 import Brick.Forms (formState, invalidFields, setFieldValid, setFormFocus)
+import qualified Data.Char as C
+import qualified Data.Text as T
 import Lens.Micro (Lens', (.~), (^.))
 
 import Summoner.Text (packageToModule)
 import Summoner.Tui.Form (KitForm, SummonForm (..), getCurrentFocus, mkForm)
 import Summoner.Tui.Kit
-
-import qualified Data.Text as T
-
 
 -- | Clears the 'Text' fields by @Ctrl + d@ key combination.
 ctrlD :: KitForm e -> KitForm e
@@ -77,6 +76,9 @@ data FormError
     | CabalOrStack
     -- | At least library or executable should be selected.
     | LibOrExe
+    -- | Prelude package name should only contain letters, numbers
+    -- and hyphens.
+    | PreludePackageValid
 
 -- | Show 'FormError' to display later in TUI.
 showFormError :: FormError -> String
@@ -86,6 +88,7 @@ showFormError = \case
     LibOrExe     -> "Choose at least one: Library or Executable"
     EmptyFields fields -> "These fields must not be empty: " ++ joinFields fields
     OneWord fields -> "These fields should contain exactly one word: " ++ joinFields fields
+    PreludePackageValid -> "Prelude package should only contain letters, numbers and hyphens"
   where
     joinFields :: NonEmpty SummonForm -> String
     joinFields = intercalate ", " . mapMaybe showField . toList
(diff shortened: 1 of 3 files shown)


summoner-tui/src/Summoner/Tui/Validation.hs
```diff
@@ -105,16 +108,22 @@ showFormError = \case
...
+    PreludePackageValid -> one CustomPreludeName
 
 -- | Takes boolean value and error and returns error if predicate 'True'.
 toError :: Bool -> e -> Validation (NonEmpty e) ()
 toError p e = if p then Failure (one e) else Success ()
 
+-- | Decides whether the given text is a valid package name. Spec is here:
+-- https://www.haskell.org/cabal/users-guide/developing-packages.html#package-names-and-versions
+packageNameValid :: Text -> Bool
+packageNameValid = T.all (\c -> c == '-' || C.isAlphaNum c)
```
vrom911: I believe this function belongs to the core package, so we would be able to reuse it for CLI as well.

summoner-tui/src/Summoner/Tui/Validation.hs
```diff
@@ -182,6 +192,13 @@ validateKit dirs kit =
         (not $ kit ^. projectMeta . lib  || kit ^. projectMeta . exe)
         LibOrExe
 
+    validatePreludePackage :: Validation (NonEmpty FormError) ()
+    validatePreludePackage =
+      let packageName = kit ^. projectMeta . preludeName in
+      toError
+        (not $ T.null packageName || packageNameValid packageName)
+        PreludePackageValid
```
chshersh: Let's move this boolean flag into the variable like `isPreludePackageInvalid` and use it later :

```haskell
toError isPreludePackageInvalid PreludePackage
```

summoner-tui/src/Summoner/Tui/Validation.hs
```diff
@@ -105,16 +108,22 @@ showFormError = \case
...
-    OneWord fields     -> fields
-    ProjectExist       -> one ProjectName
-    CabalOrStack       -> CabalField :| [StackField]
-    LibOrExe           -> Lib :| [Exe]
+    EmptyFields fields  -> fields
+    OneWord fields      -> fields
+    ProjectExist        -> one ProjectName
+    CabalOrStack        -> CabalField :| [StackField]
+    LibOrExe            -> Lib :| [Exe]
+    PreludePackageValid -> one CustomPreludeName
```
chshersh: Let's rename this constructor. You can that we don't add suffixes likes `Invalid` or `Valid` to the constructors, so more like an indicator or the check name. And semantics are attached later via boolean functions.
```suggestion
    PreludePackage -> one CustomPreludeName
```

summoner-tui/src/Summoner/Tui/Validation.hs
yigitozkavci: Hmm, I don't think null check for package name should be externalized.

We only allow empty packages because we assume that an empty package name means "no preference". Empty package names are not valid in general.

Ideal scenario would be to make it explicit that we have either specified a custom prelude package or have not, using a sum type like:

```haskell
data PreludePreference = DefaultPrelude | CustomPrelude Text
```

Also I would prefer keeping predicates as positive (ie. using `not . isPreludePackageValid` instead of `isPreludePackageInvalid`).

What do you think?

summoner-tui/src/Summoner/Tui/Validation.hs
yigitozkavci: Thanks! Now also the diff is cleaner :)

summoner-tui/src/Summoner/Tui/Validation.hs
chshersh: I find the following code more readable than the current one:

```haskell
let packageName = ...
let isPreludePackageInvalid = ...
toError isPreludePackageInvalid PreludePackage
```

It's also shorter. Because with current code `toError` application requires to split the code into multiple lines. It's hard to deduce the meaning from a bunch of boolean operators. So it makes the code easier to read if all these operators are put into a single variable with assigned meaning.

Semantically there's no difference between `isPreludePackageInvalid` and `(not isPreludePackageValid)` but if a new variable is going to be introduced, why not put this `not` application directly into variable definition :slightly_smiling_face:

yigitozkavci:
Hi, sorry for the delay here. I've fixed the conflicts and externalized the validation function. Please take another look.

summoner-tui/src/Summoner/Tui/Validation.hs
```diff
@@ -189,6 +193,14 @@ validateKit dirs kit =
         (not $ kit ^. projectMeta . lib  || kit ^. projectMeta . exe)
         LibOrExe
 
+    validatePreludePackage :: Validation (NonEmpty FormError) ()
+    validatePreludePackage =
+      let packageName = kit ^. projectMeta . preludeName in
+      let isPreludePackageInvalid = not $ T.null packageName || packageNameValid packageName in
```
vrom911: Why is there a check on `null packageName`? It is a valid option to not specify the custom prelude in the current design.

yigitozkavci: That predicate is negated, we simply allow empty and well-formatted package names here. 
`packageNameValid` function is a general-purpose check for validity of a given package name, as per Cabal spec. Since empty package names are not in general valid, but only for our application, I've localized null check here.

vrom911: Oh, yeah, sorry, it's my mistake, didn't look carefully!

yigitozkavci:
Thank you for the review! @vrom911

PR #356, opened 2019-10-25
Validate package and module name in config

vrom911:
Part of #354 

Should be tested carefully...


summoner-cli/src/Summoner/CustomPrelude.hs
```diff
@@ -0,0 +1,32 @@
+module Summoner.CustomPrelude
```
chshersh: Good idea to move this into the separate module!

summoner-cli/src/Summoner/Text.hs
```diff
@@ -22,6 +23,18 @@ packageToModule = tconcatMap headToUpper . T.splitOn "-"
...
+
+@
+Part1[.PartN]
+@
+-}
+moduleNameValid :: Text -> Bool
+moduleNameValid = all (\s -> s /= "" && validFragment s) . T.split (== '.')
+  where
+    validFragment :: Text -> Bool
+    validFragment s = T.all C.isAlphaNum s && C.isUpper (T.head s)
```
chshersh: I think it should be `T.strip s /= ""` just in case to guard from more invalid inputs... Also, what do you think about moving this this `/= ""` check to the `validFragmet` function so on the top-level this can be just?

```haskell
all isValidFragment . T.split ...
```

summoner-cli/test/Test/CustomPrelude.hs
```diff
@@ -0,0 +1,35 @@
...
+        it "no trailing dot" $
+            "Module1.Fragment2." `shouldNotSatisfy` moduleNameValid
+        it "no double dots" $
+            "Module1..Fragment2" `shouldNotSatisfy` moduleNameValid
+        it "no lower leading letter" $
+            "Module1.fragment2" `shouldNotSatisfy` moduleNameValid
+        it "no other Symbols" $
+            "Module1.fr#agment2" `shouldNotSatisfy` moduleNameValid
+        it "no several words" $
+            "Module1.fra gment2" `shouldNotSatisfy` moduleNameValid
```
chshersh: These tests are super cool! :heart_eyes_cat:

summoner-cli/src/Summoner/Text.hs
vrom911: I thought that the case with spaces should be caught in `C.alphaNum`...

Hmm, interesting idea, can move it there!

chshersh: Hmm, indeed. Makes sense!

PR #369, opened 2019-11-04
[#333] Use key-value map for extra files

### LINKED ISSUE #333: Use key-value map for extra files
Currently, we've hardcoded some of the files in the `.toml` config, for example:

```haskell
stylish.url = "https://raw.githubusercontent.com/kowainik/org/master/.stylish-haskell.yaml"
contributing.url = "https://raw.githubusercontent.com/kowainik/org/master/CONTRIBUTING.md"
```

This approach is not extensible because:

1. It doesn't allow `summoner` users to bring their own files.
2. It requires to change Haskell code each time we want to add new files.

Since `tomland-1.1` we now have the support for inline tables. And after [this issue](https://github.com/kowainik/tomland/issues/203) is implemented we can have only single field in the config of the type `Map FilePath Source` and in the configuration it will look like this:

```haskell
extra.files =
    [ { file = ".stylish-haskell.yaml", url = "https://raw.githubusercontent.com/kowainik/org/master/.stylish-haskell.yaml" }
    , { file = "CONTRIBUTING.md", url = "https://raw.githubusercontent.com/kowainik/org/master/CONTRIBUTING.md" }
    ]
``` 

What do you think?

The implementation includes:

- [x] Add `Map FilePath Source` field to config
- [x] Create files from the `Map`
- [x] Write deprecation messages
- [x] Pretty tree display in TUI

### LINKED ISSUE #360: Use `mixisns` with `relude-0.6.0.0`
- [ ] Bump up version
- [ ] Update HLint on CI

Also, I've noticed that `relude` is not inside `common-options`. I checked the code, and it looks like it should be possible to move `base-noprelude` and `relude` into `common-options`.

chshersh:
Resolves #333
Resolves #360 

It actually works! I tested with the file from `org` repo. And the tree displays perfectly! Also, `Summoner.Shell` module turned out to be redundant...


summoner-cli/src/Summoner/Project.hs
```diff
@@ -197,16 +197,26 @@ generateProject isOffline projectName Config{..} = do
...
-        infoMessage $ "Creating extra file: " <> toText path
-        whenJustM (fetchSource isOffline source) (createFileWithParents path)
-
     when settingsGitHub $ doGithubCommands settings
     beautyPrint [bold, setColor Green] "\nJob's done\n"
 
+{- | This function fetches contents of extra file sources.
+-}
+fetchSources :: Bool -> Map FilePath Source -> IO [TreeFs]
+fetchSources isOffline = mapMaybeM sourceToTree . Map.toList
```
vrom911: :sparkler: `mapMaybeM` :sparkler:

summoner-cli/src/Summoner/Project.hs
```diff
@@ -197,16 +197,26 @@ generateProject isOffline projectName Config{..} = do
...
+  where
+    sourceToTree :: (FilePath, Source) -> IO (Maybe TreeFs)
+    sourceToTree (path, source) = do
+        infoMessage $ "Fetching content of the extra file: " <> toText path
+        fetchSource isOffline source >>= \case
+            Nothing -> do
+                errorMessage $ "Error fetching: " <> toText path
+                pure Nothing
```
vrom911: Is this correct, that in case of download (or just file getting) failure the process of the project generation will still continue without those files?

summoner-cli/src/Summoner/Tree.hs
```diff
@@ -23,10 +26,52 @@ data TreeFs
...
+2. It overrides existing 'File' with the given 'TreeFs' in case of duplicates.
+-}
+insertTree :: TreeFs -> [TreeFs] -> [TreeFs]
+insertTree node [] = [node]
+insertTree node (x:xs) = case (node, x) of
+    (Dir _ _, File _ _) -> x : insertTree node xs
+    (File _ _, Dir _ _) -> x : insertTree node xs
+    (File nodePath _, File curPath _)
+        | nodePath == curPath -> node : xs
+        | otherwise -> x : insertTree node xs
```
vrom911: I don't quite get this case. When this file is already in the tree, it inserts it, but if it's not, it somehow continues inserting on the next layer? How so?

stack.yaml
```diff
@@ -6,6 +6,7 @@ packages:
 
 extra-deps:
   - optparse-applicative-0.15.0.0
+  - relude-0.6.0.0
```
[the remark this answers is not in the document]
chshersh: I've updated the boundaries in the `.cabal` file. I decided to bump up the lower bound because I'm using `mapMaybeM` function from a new version (it's very helpful :tada:) and `summon` simply won't compile with `relude-0.5.0` :disappointed: In general I don't mind supporting both versions but here I had to increase lower bound...

summoner-cli/src/Summoner/Project.hs
chshersh: @vrom911 I'm not sure, but implementing different logic is quite complicated. For example, if `isOffile` is `True` then all `Url` will fail. But maybe this is intentional and maybe I don't want things like `CONTRIBUTING.md` while developing locally :thinking:

summoner-cli/src/Summoner/Tree.hs
chshersh: Right, the implementation is not obvious, but I tried to do my best. This function always inserts `node` to the list, at least when it reaches the empty list at the end:

```haskell
insertTree node [] = [node]
```

In this case, if this function finds two `File`s with the same path, it will replace current `x` with `node` that why it's `node : xs`. This is done if people want to override default files like `.travis.yml` or `Prelude.hs` with their files. If file paths are not equal, then we keep `x`, and we will try to insert `node` into `xs`. Eventually, we either reach the end of the list, and we will add `node` unconditionally, or we will find a `File` that has the same name and will replace it.

summoner-cli/src/Summoner/Project.hs
vrom911: Yes, makes sense. By the way, where the mentioned case is handled?

summoner-cli/src/Summoner/Tree.hs
vrom911: Hmm, but are other nodes placed unconditionally? I thought there is some kind of hierarchy so no need to go till the end of the list each time.

summoner-cli/src/Summoner/Project.hs
chshersh: It's handled here:

https://github.com/kowainik/summoner/blob/38d1fceda39b8643c2860605a3eb175e387cd497/summoner-cli/src/Summoner/Source.hs#L65-L67

summoner-cli/src/Summoner/Tree.hs
chshersh: Well, this list is not sorted at this stage. It could be `[File "foo" x, File "bar" y]` or `[File "bar" x, File "foo" y]`. I checked the implementation, and it looks like we sort this list only when printing to terminal or in TUI.

Now, after you told this, I started to think that probably there's a better way to implement `TreeFs`. At least `TreeFs` shouldn't have duplicate filepaths but since we're using lists, it's possible to have a bug in the implementation and introduce duplicate fields.

Maybe instead of

```haskell
data TreeFs
    = Dir FilePath [TreeFs]
    | File FilePath Text
```

we should have

```haskell
data TreeFs
    = Dir (Map FilePath TreeFs)
    | File Text
```

The latter representation won't allow having duplicate filepaths. But I believe that the current algorithm is also correct, even if the logic is not straightforward... And changing this type is a big rewrite...

vrom911: Thank you for the clarification, Dmitrii! Yeah, I don't think that rewrite is the right step right now, and we could keep this function as it is for this PR. But let's think more about this sorting before option and create the issue if we will decide that it could work better and would become easier to maintain.

summoner-cli/src/Summoner/Project.hs
vrom911: Oh, right! Thank you :wine_glass:

summoner-cli/src/Summoner/Tree.hs
chshersh: I've created the issue:

* https://github.com/kowainik/summoner/issues/371

PR #372, opened 2019-11-05
[#333] Use key-value map for extra files

### LINKED ISSUE #333: Use key-value map for extra files

### LINKED ISSUE #360: Use `mixisns` with `relude-0.6.0.0`

chshersh:
Resolves #333


summoner-cli/src/Summoner/Config.hs
```diff
@@ -59,12 +61,13 @@ data ConfigP (p :: Phase) = Config
     , cBench        :: !Decision
     , cPrelude      :: !(Last CustomPrelude)
     , cExtensions   :: ![Text]
-    , cWarnings     :: ![Text]
-    , cGhcOptions   :: ![Text]  -- ^ GHC options to add every stanza
+    , cWarnings     :: ![Text]  -- ^ DEPRECATED: TODO: remove in 1.4.
```
vrom911: Created #374 to track this

chshersh: Thanks, that's very helpful! It's indeed a good idea to create issues to have a smoother deprecation period. Nice!

chshersh:
@vrom911 It's not a problem to update this PR, I will do this :slightly_smiling_face: Just to finish with this issue here :slightly_smiling_face:

PR #379, opened 2019-11-06
[#367] Add `Raw` constructor to `Source` data type

### LINKED ISSUE #367: Add `Content` constructor to `Source` data type
So you can specify full content of the file. This is especially nice since TOML supports multi-line strings.

chshersh:
Resolves #367

I've tested locally and it works surprisingly! If I add the following entries to the TOML configuration file (it works even if I already have `files = [ { ... } ]` entry):

```toml
[[files]]
path = "src/MyLib.hs"
raw = '''
module MyLib where

myFunc :: IO ()
myFunc = putStrLn "Hello world!"
'''

[[files]]
path = "test/Doctest.hs"
raw = '''
module Main (main) where

import System.FilePath.Glob (glob)
import Test.DocTest (doctest)


main :: IO ()
main = do
    sourceFiles <- glob "src/**/*.hs"
    doctest sourceFiles
'''
```

I can even see the files in the TUI structure. And I verifed content after creating the project, the content is perfect :ok_hand: 

Screenshot from 2019-11-06 14-29-29

## Checklist:

* Update CHANGELOG


summoner-cli/src/Summoner/Source.hs
```diff
@@ -20,19 +20,31 @@ import qualified Toml
...
-    -- | File path to the local source file.
+
+    {- | File path to the local source file.
+    -}
     | File !FilePath
-    -- | Link to external file.
+
+    {- | Link to external file. Generated file contains some short text and a
+    a give url link.
+    -}
```
vrom911: Now when we have `Raw`, this one is unnecessary. Do you think we can get rid of it?

chshersh: @vrom911 Yes, I also got the same idea :slightly_smiling_face: `Link` constructor now becomes redundant because is a strict and less useful subset of `Raw` which gives more flexibility. I think it's okay to remove it now and just write a migration guide how to use `Raw` :slightly_smiling_face:

Issue #402, opened 2020-02-17
Fix nix build


antislava:
Dependencies listed in `default.nix` are out of date and not consistent with other dependencies apparently. Also symlinks to (the new?) CHANGELOG.md file are not taken care of causing error on `nix build`.

PR #403, opened 2020-02-17
Fixing nix build 

antislava:
See [Fix nix build · Issue #402 · kowainik/summoner](https://github.com/kowainik/summoner/issues/402)

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,6 +265,12 @@ Build and install using `stack`:
 stack install summoner-tui:exe:summon-tui
 ```
 
+Build and install using `nix`:
+
+```shell
+nix-env --install --file default.nix -A summoner-tui
+```
+
 #### Summon-CLI [↑](#structure)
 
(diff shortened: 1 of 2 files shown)


antislava:
I see... (It is your call of course!) You are aware you you had these lines in the README already, aren't you?
```sh
nix-build -A summoner     # To build the CLI only
nix-build -A summoner-tui # To build the TUI
```
(which actually made me think the that `nix` installation commands were just missing). You may have to consider removing them as well in that case?

vrom911:
As I mentioned before, all `nix`-related stuff is added by the contributors. They are not maintaining this work though. So I am against adding even more stuff that we can't support.
And yes, maybe we will reconsider having these lines in the README (cc @chshersh). But I don't think that this affects my comments here and they can still be taken into consideration, can't they?

vrom911:
`Nix` stuff is now removed until we are confident enough that we can maintain that.
Closing as outdated.

PR #408, opened 2020-02-20
[#246] Put common parts of the project inside `common` stanza

### LINKED ISSUE #246: Put common parts of the project inside `common` stanza
We have several things inside our project that can be put inside `common` stanza in `.cabal` file. This will decrease the size of generated `.cabal` file significantly by removing duplication and will help people to learn new `cabal` features. Unfortunately, this is available only for `cabal-2.4`. But looks like `stack` already supports this cabal version, so there shouldn't be any major problems (though, this should be verified).

Inspired by the following `cabal` issue:

* https://github.com/haskell/cabal/issues/5705

What we can put inside `common` stanza:

1. `default-language: Haskell2010`
2. `build-depends:` — `base` or `base-noprelude` with version
3. Alternative prelude library if used
4. Default warnings
5. Default extensions

vrom911:
Resolves #246

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ To start using Summoner make sure that you have the following tools installed on
 
 We also have minimal version requirements for build tools:
 
-* [Cabal >= 2.2](https://www.haskell.org/cabal/)
+* [Cabal >= 2.4](https://www.haskell.org/cabal/)
 * [Stack >= 2.1](http://haskellstack.org)
 
 However, it is always recommended to use the newest versions of build tools.
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ Features related to the structure and content of the generated projects.
 
 + Support for Cabal and Stack build tools.
 + Ability to pick stanzas (library, executable, test-suite, benchmark).
++ Usage of [common stanza](https://vrom911.github.io/blog/common-stanzas) to reduce `.cabal` file's boilerplate.
 + Option to include an alternative prelude, if desired. The project would then
   use [`base-noprelude` technique](http://hackage.haskell.org/package/Prelude),
   and the `Prelude` module would be added to the library target.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -189,7 +190,7 @@ To start using Summoner make sure that you have the following tools installed on
 
 We also have minimal version requirements for build tools:
 
-* [Cabal >= 2.0](https://www.haskell.org/cabal/)
+* [Cabal >= 2.2](https://www.haskell.org/cabal/)
```
chshersh: Should it be `2.4`?
```suggestion
* [Cabal >= 2.4](https://www.haskell.org/cabal/)
```

README.md
```diff
@@ -84,6 +84,7 @@ Features related to the structure and content of the generated projects.
 
 + Support for Cabal and Stack build tools.
 + Ability to pick stanzas (library, executable, test-suite, benchmark).
++ Usage of [common stanza](https://vrom911.github.io/blog/common-stanzas) to reduce `.cabal` file's boilerplate.
```
chshersh: Very nice! 😍

summoner-cli/src/Summoner/Template/Cabal.hs
```diff
@@ -93,85 +109,65 @@ cabalFile Settings{..} = File (toString settingsRepo ++ ".cabal") cabalFileConte
         [text|
         $endLine
         library
+          import:              common-options
           hs-source-dirs:      src
           exposed-modules:     $libModuleName
                                $preludeMod
+        |]
```
chshersh: I like how even our code becomes simpler because of common stanzas!

README.md
vrom911: Yeah, let's not overcomplicate things and put the one we are currently using :)

PR #409, opened 2020-02-20
[#361] Put all warnings under `if`s to always have them all

### LINKED ISSUE #361: Put all warnings under `if`s to always have them all
Blocked by #246 

Once we put all warnings into common stanzas, we can always put all GHC options and warnings under `if` conditionals. This is more useful because we can automatically see all warnings and always have them even when the project is updated.

chshersh:
Resolves #361


CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ The changelog is available [on GitHub][2].
 * [#363](https://github.com/kowainik/summoner/issues/363):
   Move from `generic-deriving` to `generic-data`.
   (by [@chshersh](https://github.com/chshersh))
+* [#361](https://github.com/kowainik/summoner/issues/361):
+  Always put all warnings in `ghc-options` inside common stanza under `if`s.
```
vrom911: Let's add an example of `if` in here if you don't mind :slightly_smiling_face: 
Also, maybe makes sense to mention that the condition is on GHC in these `if`s? Or only one of these two options is fine as well, what do you think?

summoner-cli/src/Summoner/License.hs
```diff
@@ -65,11 +64,6 @@ newtype License = License
 instance FromJSON License where
     parseJSON = withObject "License" $ \o -> License <$> o .: "body"
 
--- | As it will be shown in the @cabal@ file.
-cabalLicense :: LicenseName -> Text
-cabalLicense None = "AllRightsReserved"
-cabalLicense l    = show l
-
```
vrom911: That is a nice simplification!

summoner-cli/src/Summoner/Template/Cabal.hs
```diff
@@ -95,15 +94,42 @@ cabalFile Settings{..} = File (toString settingsRepo ++ ".cabal") cabalFileConte
...
     |] <> defaultExtensions
 
     baseBounds :: Text
     baseBounds = cabalBaseVersions settingsTestedVersions
 
+    ghcOptions :: Text
+    ghcOptions = case settingsGhcOptions of
+        [] -> defaultGhcOptions
+        x:xs ->
+            let customGhcOptions = T.intercalate "\n" $ x : map (T.replicate 21 " " <>) xs in
```
vrom911: Could `intercalateMap` function be handy in here?

chshersh: Yes, it can help here indeed! I missed that

vrom911: Or not.. looking into it I see, that the logic here is a bit different. Sorry, false alarm!

PR #420, opened 2020-02-23
[#70] Add 'summon config' command

### LINKED ISSUE #70: Add 'summon config' command
This command should create `.summoner.toml` file in interactive, user-friendly way. Also, probably it can have better name than `summon init`.

chshersh:
Resolves #70

I appreciate any suggestions on the helper text for the config :slightly_smiling_face:


summoner-cli/test/Test/TomlSpec.hs
```diff
@@ -18,6 +22,16 @@ import qualified Hedgehog.Gen as Gen
 import qualified Hedgehog.Range as Range
 
 
+tomlSpec :: Spec
+tomlSpec = describe "TOML configuration spec" $ do
+    it "finalises default configuration" $
+        finalise defaultConfig `shouldSatisfy` isSuccess
+    it "parses default configuration" $
+        decode configCodec defaultConfigFileContent `shouldSatisfy` isRight
```
chshersh: This is the best I could come up in terms of testing. Ideally, it would be nice to parse with uncommented fields, but I can't just drop `#` in front of each line since some text is the help text.

summoner-cli/src/Summoner/CLI.hs
```diff
@@ -85,6 +86,35 @@ runCliCommand = \case
...
+    alreadyExist <- doesFileExist configFile
+    if alreadyExist
+    then do
+        warningMessage $ "File '" <> configFileTxt <> "' already exits."
+        infoMessage "Use 'summon config --file <path>' to specify another path."
+        exitFailure
+    else do
+        writeFileText configFile defaultConfigFileContent
+        infoMessage $ "Created default configuration file: " <> configFileTxt
+        infoMessage "Open the file using your editor of your choice."
```
vrom911:
```suggestion
        infoMessage "Open the file using the editor of your choice."
```

summoner-cli/src/Summoner/Default.hs
```diff
@@ -44,3 +65,103 @@ currentYear = do
...
+# test = true
+
+# Should your projects have the benchmark stanza by default?
+# bench = true
+
+# Summoner suports 'git' version control system and GitHub as a platform for
+# hosting 'git' repos. The following set of flags controls integration with GitHub.
+# github = true    # enabled GitHub integration
+# noUpload = true  # Init git repo, but don't create it on GitHub
+# private = false  # create private repos by default
```
vrom911:
```suggestion
# private = true  # create private repos by default
```

summoner-cli/src/Summoner/Default.hs
```diff
@@ -44,3 +65,103 @@ currentYear = do
...
+
+# List of additional files to add after creating the project
+# files =
+#     [ { path = ".stylish-haskell.yaml"
+#       , url  = "https://raw.githubusercontent.com/kowainik/org/master/.stylish-haskell.yaml"
+#       }
+#     , { path = ".github/CODEOWNERS"
+#       , raw  = "*  @$defaultOwner"
+#       }
+#     , { path  = ".github/CODEOWNERS"
```
vrom911:
```suggestion
#     , { path  = "src/Foo.hs"
```

summoner-cli/test/Test/TomlSpec.hs
vrom911: I see.
I think golden testing I described above could help with the issue :slightly_smiling_face:

PR #425, opened 2020-02-25
Better GitHub releases

chshersh:
This PR is an attempt to bring better binary GitHub releases to `summoner`.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,15 +31,14 @@ build_script:
   - stack build --test --no-run-tests --no-run-benchmarks
   # prepare binary releases
   - stack install summoner --local-bin-path bin
-  - if /I "%APPVEYOR_REPO_TAG%" EQU "true" (set DEPLOY_TAG=%APPVEYOR_REPO_TAG_NAME%-x86_64-windows.zip)
   - set DEPLOY_DIR=%APPVEYOR_BUILD_FOLDER%\bin
-  - if /I "%DEPLOY_TAG%" NEQ "" (7z a "bin\summoner-%DEPLOY_TAG%" "%DEPLOY_DIR%\summoner.exe")
+  - if /I "%DEPLOY_TAG%" NEQ "" (7z a "bin\summoner-cli-x86_64-windows.zip" "%DEPLOY_DIR%\summoner.exe")
 
 test_script:
   - stack test
 
 artifacts:
-  - path: bin\summoner-%DEPLOY_TAG%
+  - path: bin\summoner-cli-x86_64-windows.zip
     name: summoner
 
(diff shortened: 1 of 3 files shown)


chshersh:
Okay, so apparently, I can't uncomment macOS job easily. I see two possible solutions to proceed with this:

1. Use only `stack` on Travis CI to produce binaries for both mac and Linux.
2. Investigate [GitHub Actions releases](https://github.com/andys8/git-brunch/blob/master/.github/workflows/release.yml)

@vrom911 what do you think?

vrom911:
@chshersh I don't mind having only Stack at one CI platform as long as we have the Cabal check somewhere else. I hope it will still produce good enough binaries :slightly_smiling_face: 

GitHub actions look promising, we can look into that, but only if it's not too time-consuming. In case of that, we can just go with the Stack-only option. What you personally prefer?

chshersh:
@vrom911 I never used GitHub Actions for releases before. From what I saw, they don't look too difificult, but maybe there's a hidden catch. The problem is that to produce binaries, it's easier to use `stack` since `cabal` doesn't have CLI to get paths to produced binaries. But I don't know how to have `stack` on GitHub actions and have fast cached builds as well.

I'm also okay with using only `stack` at Travis CI, since we have `cabal` on GitHub actions 🙂 And AppVeyor CI also will have only `stack`...

chshersh:
This PR should be ready to merge once CI passes. To highlight changes:

1. We now have better GitHub releases for all 3 operating systems. You can see example of such release here:
    * https://github.com/kowainik/summoner/releases/tag/v2.0-draft
    Any suggestions on names for files in there are appreciated 🙂 
2. We have robust Travis CI for both linux and macOS using `stack`.
3. AppVeyor was switched to `stack` as well.
4. All cabal example projects are tested properly with `cabal build` on GItHub actions.
5. All stack example projects are tested properly with `stack build` on Travis CI

appveyor.yml
```diff
@@ -8,27 +7,51 @@ platform:
...
 
 environment:
-  global:
-    CABOPTS: --store-dir=C:\\SR
+  STACK_ROOT: C:\sr
+  STACK_VERSION: 2.1.1
+
+  # Workaround a gnarly bug https://github.com/haskell/cabal/issues/5386
+  # See: https://www.fpcomplete.com/blog/2018/06/sed-a-debugging-story
+  # TODO: check if it's fixed once we switch to lst-13 and GHC 8.6
```
vrom911: That might be an outdated comment :slightly_smiling_face:

chshersh: @vrom911 Good point! I'll remove it 👍

PR #426, opened 2020-02-28
Typo in default config

steshaw:


steshaw:
Oops. That should do it.

I found the golden file at `summoner-cli/examples/summoner-default.toml`.

steshaw:
No worries!

PR #431, opened 2020-03-17
[#429] Move to GHC-8.8.3

### LINKED ISSUE #429: Move to GHC-8.8.3
Stackage lts-15.3 with GHC-8..8.3 is out :tada: 
So we can use it for the generated projects :+1:

vrom911:
Resolves #429

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-15.0
+resolver: lts-15.3
 
 packages:
   - summoner-cli/
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-15.3
+resolver: lts-15.4
 
 packages:
   - summoner-cli/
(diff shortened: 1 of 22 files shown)


stack.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: lts-15.0
+resolver: lts-15.3
```
chshersh: Probably should be `15.4`
```suggestion
resolver: lts-15.4
```

summoner-cli/summoner.cabal
```diff
@@ -19,7 +19,7 @@ extra-doc-files:     README.md
 tested-with:         GHC == 8.2.2
                    , GHC == 8.4.4
                    , GHC == 8.6.5
-                   , GHC == 8.8.2
+                   , GHC == 8.8.3
```
chshersh: Should we also remove leading `,` in our cabal files?

PR #432, opened 2020-03-22
[#430] Allow-failures for old GHC on Stack at TravisCI

### LINKED ISSUE #430: Allow-failures for old GHC on Stack at TravisCI
Instead of just not including old GHCs for Stack check at Travis, we can use `allow_failures:` field for these matrix elements.
What do you think, @chshersh?

vrom911:
Resolves #430


vrom911:
I have tested it in this PR: https://github.com/chshersh/stack-full/pull/1

PR #457, opened 2020-05-18
Use headroom to prettify headers

chshersh:
Using [headroom](https://github.com/vaclavsvejcar/headroom) to manage headers automatically. I've performed the following steps:

1. Run `headroom init -l mpl2 -s summoner-cli/src/`
2. Manually edit `headroom` files
3. Run `headroom run -v "author=Kowainik" -v "year=2020"`

It's a nice tool :slightly_smiling_face: But there are a few things that stop us from using `headroom` in all our projects. Maybe my configuration is wrong in some places, so I would appreciate any tips.

1. Currently `headroom` doesn't replace the top-level module documentation, it always adds a new entry (even when `replace` mode is enabled)
2. It doesn't substitute `LONG_DESC` and `MODULE_NAME` variables.
3. It doesn't preserve existing years (since it doesn't replace the header) and doesn't update years automatically (but there are open issues about that).

I'm opening the PR so we can discuss possible configuration adjustments and the possibility to use `headroom` in our projects :slightly_smiling_face: 

cc @vaclavsvejcar

### DIFF
--- a/summoner-cli/src/Summoner.hs
+++ b/summoner-cli/src/Summoner.hs
@@ -1,7 +1,10 @@
 {- |
-Copyright: (c) 2017-2019 Kowainik
-SPDX-License-Identifier: MPL-2.0
-Maintainer: Kowainik <xrom.xkov@gmail.com>
+Module                  : Summoner
+Copyright               : (c) 2017-2020 Kowainik
+SPDX-License-Identifier : MPL-2.0
+Maintainer              : Kowainik <xrom.xkov@gmail.com>
+Stability               : Stable
+Portability             : Portable
 
 Main module that reexports all library components of the @summoner@.
 -}
--- a/summoner-cli/src/Summoner/CLI.hs
+++ b/summoner-cli/src/Summoner/CLI.hs
@@ -4,9 +4,12 @@
 {-# LANGUAGE TupleSections   #-}
 
 {- |
-Copyright: (c) 2017-2019 Kowainik
-SPDX-License-Identifier: MPL-2.0
-Maintainer: Kowainik <xrom.xkov@gmail.com>
+Module                  : Summoner.CLI
+Copyright               : (c) 2017-2020 Kowainik
+SPDX-License-Identifier : MPL-2.0
+Maintainer              : Kowainik <xrom.xkov@gmail.com>
+Stability               : Stable
+Portability             : Portable
 
 This module contains functions and data types to parse CLI inputs.
 -}
--- a/summoner-cli/src/Summoner/Config.hs
+++ b/summoner-cli/src/Summoner/Config.hs
@@ -7,9 +7,12 @@
 {-# LANGUAGE UndecidableInstances #-}
 
 {- |
-Copyright: (c) 2017-2019 Kowainik
-SPDX-License-Identifier: MPL-2.0
-Maintainer: Kowainik <xrom.xkov@gmail.com>
+Module                  : Summoner.Config
+Copyright               : (c) 2017-2020 Kowainik
+SPDX-License-Identifier : MPL-2.0
+Maintainer              : Kowainik <xrom.xkov@gmail.com>
+Stability               : Stable
+Portability             : Portable
 
 Summoner configurations.
 -}
--- a/summoner-cli/src/Summoner/CustomPrelude.hs
+++ b/summoner-cli/src/Summoner/CustomPrelude.hs
@@ -1,7 +1,10 @@
 {- |
-Copyright: (c) 2017-2019 Kowainik
-SPDX-License-Identifier: MPL-2.0
-Maintainer: Kowainik <xrom.xkov@gmail.com>
+Module                  : Summoner.CustomPrelude
+Copyright               : (c) 2017-2020 Kowainik
+SPDX-License-Identifier : MPL-2.0
+Maintainer              : Kowainik <xrom.xkov@gmail.com>
+Stability               : Stable
+Portability             : Portable
 
 This module implements data type for representing custom alternative preludes.
 -}
--- a/summoner-cli/src/Summoner/Decision.hs
+++ b/summoner-cli/src/Summoner/Decision.hs
@@ -1,7 +1,10 @@
 {- |
-Copyright: (c) 2017-2019 Kowainik
-SPDX-License-Identifier: MPL-2.0
-Maintainer: Kowainik <xrom.xkov@gmail.com>
+Module                  : Summoner.Decision
+Copyright               : (c) 2017-2020 Kowainik
+SPDX-License-Identifier : MPL-2.0
+Maintainer              : Kowainik <xrom.xkov@gmail.com>
+Stability               : Stable
+Portability             : Portable
 
 Decision data type.
 -}
--- a/summoner-cli/src/Summoner/Default.hs
+++ b/summoner-cli/src/Summoner/Default.hs
@@ -1,9 +1,12 @@
 {-# LANGUAGE QuasiQuotes #-}
 
 {- |
-Copyright: (c) 2017-2019 Kowainik
-SPDX-License-Identifier: MPL-2.0
-Maintainer: Kowainik <xrom.xkov@gmail.com>
+Module                  : Summoner.Default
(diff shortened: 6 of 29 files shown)


headroom-templates/haskell.mustache
```diff
@@ -0,0 +1,10 @@
+{- |
+Module                  : MODULE_NAME
```
vrom911, line 2: Maybe it should be in double brackets too? I am not an expert, but usually, in mustache all the variables should be in such brackets.

chshersh: @vrom911 Makes sense to me. It seems like a reasonable UI where all variables are put in `{{ }}` and some of the variables are just provided by `headroom`.

vaclavsvejcar:
Let's start with the issue from top:

#### Ad 1
I think the problem is that in `.headroom.yaml` you configured that license headers will use the _single-line_ style comments: https://github.com/kowainik/summoner/blob/6ef4e97cf5efa2fcaadaecb2f24aea9c73346d17/.headroom.yaml#L20
but the [template you use](https://github.com/kowainik/summoner/blob/6ef4e97cf5efa2fcaadaecb2f24aea9c73346d17/headroom-templates/haskell.mustache) uses block comments (`{- |` and ` -}`). When _Headroom_ looks for already existing headers to replace, it reads the configuration to know how the license header will look like, so it expects existing license header that uses _single line comment_. That is the reason why it ignores the existing _block-comment_ header and adds another one to your source code files. To fix this, it should be enough edit the configuration to use proper _block-comment_ style syntax. So if you want to use _block-comment_ syntax with 1 blank line of margin before and after the header and to put the header after language pragmas but before the `module Foo where ...` line, you can use something like this:
```yaml
license-headers:
  haskell:
    put-after: ["^{-#"]
    margin-after: 1
    margin-before: 1
    block-comment:
      starts-with: "{- |"
      ends-with: "-}"
```

The explicit `block-comment` is necessary if you want your block comment to start with `{- |`, as the default value in _Headroom_ is `{-|` (without the extra space).

Also please let me know if this need of consistency between configuration and actual content of `.mustache` template isn't clear from the _Headroom_ documentation, because in that case it's obviously failure of the documentation to properly explain this.

Before trying these changes, you might want to rollback the changes you already made, as there are already two module headers present in the code, or you can try to drop them using `headroom run -d -v "author=Kowainik" -v "year=2020"` and then re-running `headroom run -d -v "author=Kowainik" -v "year=2020"` to add the new ones back (yup I know that it's annoying to define variables and template paths for dropping mode, it's bug and it will be fixed here: https://github.com/vaclavsvejcar/headroom/issues/50). But we'll need to also solve **2)** and **3)** to make the headers generated by _Headroom_ useful :-)

headroom-templates/haskell.mustache
vaclavsvejcar: I think we're getting to point **2)** from above - Problem is that currently _Headroom_ cannot extract any variable values from Haddock module header already existing in processed source code file. Now, the only source of variable values is either `variables` section in `.headroom.yaml` or `-v|--variable "KEY=VALUE"` command line argument. There is existing issue https://github.com/vaclavsvejcar/headroom/issues/25 that should make this possible. Simply said, _Headroom_ will contain custom code for every supported file type that will be able to extract some specific content from source code files for templates - such as to parse existing Haddock module header into some variable values for _Haskell_ source code files. This is becoming top priority for next release, as it bothers me too, I just haven't had enough time to think about it thoroughly and implement it.

vaclavsvejcar:
@chshersh @vrom911 For last couple of weeks I was implementing all the required stuff to hopefully make _Headroom_ more usable for you. All new changes are in the `master` branch of _Headroom_ repo, this version hasn't been released as public yet. To make things easier, I forked the _Summoner_ and pushed proposed _Headroom_ configuration that should work for you - https://github.com/vaclavsvejcar/summoner - it seems to work mostly fine, just needs manual intervention in couple of files, details below.

Snímek obrazovky 2020-06-19 v 21 00 32

### License Template
See https://github.com/vaclavsvejcar/summoner/blob/master/headroom-templates/haskell.mustache, variables starting with _underscore_ are the _dynamic variable_, defined for each parsed file, for example the `_haskell_module_copyright` holds the `Copyright` field from possibly existing _Haddock_ module header from parsed file. I also used triple brackets instead of double, as that forces _Mustache_ to render content unescaped (it escapes for HTML by default).

### Configuration File
See https://github.com/vaclavsvejcar/summoner/blob/master/.headroom.yaml, what's worth mentioning:

#### Variables

```yaml
variables:
  author: Kowainik
  email: xrom.xkov@gmail.com
  _haskell_module_copyright: "(c) {{ _current_year }} {{ author }}"
```

First two variables are the _static ones_, third value is the default value for the above mentioned `_haskell_module_copyright` _dynamic variable_. This value is used in case that the parsed file doesn't contain _Haddock_ module header (like brand new file), but you still want to put some meaningful value in there. In general, variable values can be Mustache templates, as long as you avoid cyclic dependencies or recursive references.

#### Header configuration
```yaml
license-headers:
  haskell:
    put-after: ["^{-#"].   # put header after LANGUAGE pragmas
    margin-after: 1        # blank line space after
    margin-before: 1       # and before header
    block-comment:
      starts-with: "{- |". # default is '{-|' (without space) so you need to override this
      ends-with: "-}"
```

These are basically some overrides of the default configuration, let me know if you need to configure something differently.

#### Postprocessing functions
One of your requirements was to update years for current author in _Haddock_ module header. This can be done using the _postprocessing header function_ called `update-copyright`:

```yaml
post-process:
  update-copyright:
    enabled: true
    config:
      selected-authors-only: ["{{ author }}"]
```

The `selected-authors-only` allows to only update lines containing selected author names, and as you can see, you can reuse here the value of `{{ author }}` variable defined if `variables` section of _YAML_ file.

### Files that need manual changes
You can try clone my forked version of _Summoner_ and run `headroom run` to replace headers. It works mostly fine, only in following files, you'd need to probably manually change few things before running _Headroom_:

1. `Project.hs` - Headroom will unfortunately drop the `{- HLINT ignore "Reduce duplication" -}` line. It that's an issue, we can try to figure out how to convince Headroom not to touch it.
1. `GitHub.hs` - There is existing license header, but starts with `{-|` (without the space between `-` and `|`), which is not considered as existing header by _Headroom_ and new header is placed above. Simple change from `{-|` to `{- |` should make it.

vrom911:
@vaclavsvejcar That's an incredible piece of work :heart_eyes: Thanks a lot for working on this, looks fantastic!

Regarding the files that need manual changes, I guess, the missing space is our bad, and we should change that. 
Regarding HLINT, we can use the solution that you suggested, or move such pragmas around, I see that it might help, if we place it after imports, or near the enabled HLINT code block.

vaclavsvejcar:
@vrom911 Awesome! 🙂 Let me know if you'll need some additional features or help with setup. All these changes will be part of upcoming major release `0.3.0.0`, most of the functionality should be ready, but I need to extend the documentation on the [end-user documentation microsite](https://doc.norcane.com/headroom/latest/).

vrom911:
Looking forward to the new release, you are doing amazing work there! :hugs: 

By the way, would you like to make the PR with the config for Summoner, that you kindly prepared for us here: https://github.com/vaclavsvejcar/summoner/commit/9a35ace2b89bf1da20b78e9c1f2ea15133240a99 ?
We would then rebase on that and try to apply it in this PR :slightly_smiling_face:

vaclavsvejcar:
> ...would you like to make the PR with the config for Summoner...

Sure, no problem 🙂 Should I include the configuration only and you'll play with the headers on your own, or should I run _Headroom_ & modify the headers as part of the pull request?

vrom911:
Config only sounds perfect. It is interesting to run `Headroom` on our own to actually test the settings as well 🙂

vaclavsvejcar:
@vrom911 @chshersh Here #471 is the PR with proposed _Headroom_ configuration, I already had it prepared, just not pushed, so decided to propose it to hopefully save you some time with configuration 🙂

chshersh:
I've updated license headers using `headroom` and the config contributed by @vaclavsvejcar. Looks neat!

PR #479, opened 2020-07-03
[#476] Swap aeson with microaeson

### LINKED ISSUE #476: Switch from 'aeson' to 'microaeson'
This will make `summoner` much lighter and faster to install. The only problem is that `microaeson` is not on Stackage.

willbasky:
Resolve #476

### DIFF
--- a/summoner-cli/src/Summoner/Config.hs
+++ b/summoner-cli/src/Summoner/Config.hs
@@ -32,6 +32,7 @@ module Summoner.Config
 
 import Data.List (lookup)
 import Generic.Data (gmappend, gmempty)
+import Generic.Data.Orphans ()
 import Shellmet (($?), ($|))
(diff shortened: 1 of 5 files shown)


summoner-cli/src/Summoner/Config.hs
```diff
@@ -32,6 +32,7 @@ module Summoner.Config
 
 import Data.List (lookup)
 import Generic.Data (gmappend, gmempty)
+import Generic.Data.Orphans ()
```
vrom911, line 35: Do I understand this correctly, that this is needed only for older versions of GHC?

willbasky:
Not so much, just better though.

summoner-cli/src/Summoner/Config.hs
willbasky: Yes, i found it from lib author. It didn't build 8.4.4 without orphans import.

PR #505, opened 2020-09-08
Update travis.yaml template

### LINKED ISSUE #504: Generated .travis.yaml has linter issues.
Linter issues:
image

I believe that there is no reason not to remove `sudo` and use `jobs` in place of `matrix`.  Not specifying `os` and `dist` allows automatic update in case Travis changes the defaults. `cabal` warning is false positive according to https://docs.travis-ci.com/user/languages/haskell/ .

gdziadkiewicz:
*  remove deprecated `sudo` key
*  replace matrix with jobs

Fixes #504 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: true
 language: generic
 
 git:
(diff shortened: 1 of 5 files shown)


vrom911 (approved):
Great! Thanks for fixing this 👍🏼 
I have also tested it on our test projects 🙂

Issue #340, opened 2019-08-28
Move to cabal-3.0


chshersh:
This change implies:

- [x] Changing `defaultCabal`
- [x] Removing `v2-` prefixes
- [x] Simplify `cabal test` (`echo "No tests"` no longer needed)
- [x] Add `--write-ghc-environment-files=always` option

The following files need to be changed:

- [ ] `.travis.yml`
- [ ] Generated `.travis.yml`
- [ ] `appveyor.yml`
- [ ] Generated `appveyor.yml`
- [ ] `.github/workflows/ci.yml`
- [ ] Generated `.github/workflows/ci.yml`

vrom911:
And changing cabal version in `.cabal` file and `.travis.yml`.

srid:
> Future: generated default.nix

Could you elaborate on what this means? I thought `default.nix` is hand-written?

chshersh:
@srid The word _generated_ in the context of this project means _files and directories generated by `summoner` during his work_. So ideally we are trying to use the same CI configs as `summoner` would generate during scaffolding. Dogfooding this project helps to improve UX :+1: 

After looking closely at `default.nix` again I don't see any `cabal`-related information in this package.... So we might not even need to change it at all if it uses `cabal >= 3.0` by default.

szabi:
Is Grzegorz  @gdziadkiewicz still working on this?

gdziadkiewicz:
@szabi I'm not

PR #511, opened 2020-10-11
[#454] Implement cross-platform GitHub Actions.

### LINKED ISSUE #454: Implement cross-platform GitHub Actions
Described in this blog post:

* https://kodimensional.dev/github-actions

gdziadkiewicz:
Fix #454


summoner-cli/examples/cabal-full/.github/workflows/ci.yml
```diff
@@ -7,39 +7,59 @@ on:
...
     - name: Build
       run: |
-        cabal v2-update
-        cabal v2-build --enable-tests --enable-benchmarks
+        cabal v2-configure --enable-tests --enable-benchmarks
+        cabal v2-build all
 
     - name: Test
       run: |
-        cabal v2-test --enable-tests --test-show-details=direct
```
gdziadkiewicz, line 45: `--test-show-details=direct` didn't work for me with Cabal 2.4 (see https://github.com/gdziadkiewicz/ubiquitous-octo-umbrella/runs/1237695655?check_suite_focus=true#step:6:4 ).

summoner-cli/examples/full-batteries/.github/workflows/ci.yml
```diff
@@ -20,28 +21,92 @@ jobs:
...
+        cabal v2-configure --enable-tests --enable-benchmarks
+        cabal v2-build all
+
+    - name: Test
+      run: |
+        cabal v2-test all
+
+  stack:
+    name: stack / ghc ${{ matrix.ghc }}
+    runs-on: ubuntu-latest
```
gdziadkiewicz, line 78: @chshersh I went with your approach from https://kodimensional.dev/github-actions and used only Linux, but I don't understand the reason behind it. Why not do the same thing as for cabal in terms of versions(all for Linux, latest for rest) and oses?

gdziadkiewicz:
Repo created using Summoner with this change: https://github.com/gdziadkiewicz/NowyTestowy

summoner-cli/examples/cabal-full/.github/workflows/ci.yml
chshersh: Yes, this is a known problem. This option is added only in `master`, not in the previous release. We were planning to move to Cabal-3.0 in the next release, so this will work eventually: https://github.com/kowainik/summoner/issues/340

chshersh: For now, it's okay to keep `--test-show-details` as it is :slightly_smiling_face:

summoner-cli/examples/full-batteries/.github/workflows/ci.yml
chshersh: The current approach described in my blog post optimizes for the low number of builds and rapid development cycles, so you don't need to wait hours for the CI to finish. Usually, for Haskell projects it's enough to do the following:

* Test at least once with each GHC version
* Test at least once on each OS
* Test at least once with Cabal
* Test at least once with Stack

Since testing multiple GHC versions with Stack is more awkward than with Cabal and because Stack build cache is much bigger than the Cabal cache, we prefer Cabal by default with different combinations of GHC and OS, and Stack is there only to make sure that your `stack.yaml` is up-to-date.

Yes, there are extremely rare cases when some combinations of GHC versions and OS may not work (and since GHC is broken on Windows most of the time, you will get a lot of failed CI builds on Windows). But testing everything on everything is usually a waste of time for most projects.

summoner-cli/src/Summoner/Project.hs
```diff
@@ -98,7 +98,7 @@ generateProjectInteractive connectMode projectName ConfigP{..} = do
         (settingsGitHub && not settingsNoUpload)
         (YesNoPrompt "private repository" "Create as a private repository (Requires a GitHub private repo plan)?")
         cPrivate
-    settingsGhActions <- decisionIf (settingsCabal && settingsGitHub) (mkDefaultYesNoPrompt "GitHub Actions CI integration") cGhActions
+    settingsGhActions <- decisionIf settingsGitHub (mkDefaultYesNoPrompt "GitHub Actions CI integration") cGhActions
```
chshersh, line 101: Nice catch!

summoner-cli/src/Summoner/Template/GitHub.hs
```diff
@@ -121,49 +121,113 @@ gitHubFiles Settings{..} = concat
...
         , "    strategy:"
         , "      matrix:"
+        , "        os: [ubuntu-latest, macOS-latest, windows-latest]"
         , "        cabal: [" <> quote defaultCabal <> "]"
         , "        ghc:"
         ]
         <> map (indent 10 <>) ghActionsVersions
         <>
+        [ "        exclude:"]
+        <> map (indent 10 <>) ghActionsExcludes
```
chshersh, line 142: I think, if you selected only single GHC version, there should be no `exclude` at all for `cabal-install` builds

summoner-cli/src/Summoner/Template/GitHub.hs
```diff
@@ -121,49 +121,113 @@ gitHubFiles Settings{..} = concat
...
+        <>
         [ ""
         , "    steps:"
         , "    - uses: actions/checkout@v2"
         , "      if: github.event.action == 'opened' || github.event.action == 'synchronize' || github.event.ref == 'refs/heads/master'"
         , ""
-        , "    - uses: actions/setup-haskell@v1"
+        , "    - uses: actions/setup-haskell@v1.1.1"
```
chshersh, line 149: Since we are implementing this, let's also use the latest versions of all actions at the moment. It probably won't be that easy to update, but for now we can use the latest versions. For `setup-haskell` the latest version if `1.1.3`. And, I guess, we can move all such versions into variables to make the easier to update.
```suggestion
        , "    - uses: actions/setup-haskell@v${defaultSetupHaskellVersion}"
```

summoner-cli/src/Summoner/Template/GitHub.hs
```diff
@@ -121,49 +121,113 @@ gitHubFiles Settings{..} = concat
...
         , "      name: Setup Haskell"
         , "      with:"
         , "        ghc-version: ${{ matrix.ghc }}"
         , "        cabal-version: ${{ matrix.cabal }}"
         , ""
+        , "    - name: Freeze"
+        , "      run: |"
+        , "        cabal v2-freeze"
+        , ""
         , "    - uses: actions/cache@v1"
```
chshersh, line 160: Same for the `cache` action. I guess the latest version is 2.1.2, but you can verify this :slightly_smiling_face:

summoner-cli/src/Summoner/Template/GitHub.hs
```diff
@@ -294,12 +360,15 @@ gitHubFiles Settings{..} = concat
...
+    cabalConfigure :: Text
+    cabalConfigure = "cabal v2-configure --enable-tests --enable-benchmarks"
+
     cabalBuild :: Text
-    cabalBuild = "cabal v2-build --enable-tests --enable-benchmarks"
+    cabalBuild = "cabal v2-build all"
 
     cabalTest :: Text
     cabalTest = if settingsTest
-        then "cabal v2-test --enable-tests --test-show-details=direct"
```
chshersh, line 302: This flags we implemented under #466, and the proper fix is to bump up Cabal version on CI, not delete flags. The `--test-show-details=direct` flag should go to the `configure` command

chshersh (commented):
@gdziadkiewicz Nice work! I left some comments and some things to keep in mind.

The implementation looks good :+1: But a few more patches required :slightly_smiling_face:

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: Ok, I will bump Cabal version and restore the flags

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: I will verify and update.

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: I will use the newest versions of actions and extract the versions into variables to make maintenance easier.

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: That's true - will fix that.

summoner-cli/examples/cabal-full/.github/workflows/ci.yml
gdziadkiewicz: `--test-show-details=direct` has been restored in `cabal configure` call.

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: Flag restored, Cabal bumped to 3.0.

chshersh:
Hi @gdziadkiewicz! I see that you've also moved everything to Cabal-3.0. This is a separate big issue, and ideally should be done separately. For now, the only thing required for GitHub Actions is to use `cabal 3.0` in `.github/workflows/ci.yml` file. Can you please revert the commit with all changes to Cabal 3.0 and fix only the GitHub Actions workflow? Moving everything to Cabal 3.0 is tricky and requires more thoughtful consideration and testing, and we are planning to do it after some other fixes.

gdziadkiewicz:
@chshersh I reverted cabal 3.0 changes, and now I'm working on remaining issues.

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: Done.

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: Done.

summoner-cli/src/Summoner/Template/GitHub.hs
```diff
@@ -121,49 +121,121 @@ gitHubFiles Settings{..} = concat
...
         , ""
         , "jobs:"
-        , "  build:"
-        , "    name: ghc ${{ matrix.ghc }}"
-        , "    runs-on: ubuntu-16.04"
+        ]
+        <> memptyIfFalse settingsCabal ghActionsCabal
+        <> memptyIfFalse settingsStack ghActionsStack
+
+    ghcActionsCheckoutVersion = "@v2.3.3"
```
gdziadkiewicz, line 128: Is this place ok for keeping the versions?

gdziadkiewicz:
Committed changes to https://github.com/gdziadkiewicz/NowyTestowy

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: Fixed and checked

gdziadkiewicz:
I believe that this is ready to be rereviewed

gdziadkiewicz:
@chshersh While working on GitHub Actions for my Haskell project, I found out that `actions/setup-haskell` will install unneeded Cabal and global GHC if configured the way we do it here/ in your article. I checked the documentation for it, and it confirms that this is expected. The missing part to avoid having it installed is to add:
```
stack-no-global: true
enable-stack: true
```
to the stack job.

What's your opinion on adding that?

summoner-cli/src/Summoner/Default.hs
```diff
@@ -42,6 +43,10 @@ defaultGHC = maxBound
 defaultCabal :: Text
 defaultCabal = "2.4"
 
+-- | Default version of the Stack.
+defaultStack :: Text
+defaultStack = "2.3.3"
```
gdziadkiewicz, line 48: 2.5.1 was released since I did that. Should I update it?

gdziadkiewicz: Updated to 2.5.1 .

appveyor.yml
```diff
@@ -17,7 +17,7 @@ cache:
 
 environment:
   STACK_ROOT: C:\sr
-  STACK_VERSION: 2.1.1
+  STACK_VERSION: 2.5.1
```
chshersh, line 20: I see that Stack 2.5.1 is on Chocolatey, so this should work 👍 
Just a lot of places needed to be look before bumping versions of tools everywhere...

* https://chocolatey.org/packages/haskell-stack

summoner-cli/src/Summoner/Template/GitHub.hs
chshersh: I think it's okay. As they are only used here, so no need to move to some external module.

chshersh (approved):
Looking good! 👍 I hope I didn't miss anything 🤞  Waiting for the review from @vrom911, she can review very carefully and notice anything suspicious.

> I found out that actions/setup-haskell will install unneeded Cabal and global GHC if configured the way we do it here
> What's your opinion on adding that?

Can you elaborate a bit more on that? The fact that `setup-haskell` installs global GHC is plus ➕  It can do it faster and more efficiently than Stack (either via `ghcup` or even using already provided by environment version). So, in Stack build we use the `--system-ghc` flag to tell Stack to use already installed GHC and not install it on its own.

gdziadkiewicz:
I missed the use of  `--system-ghc` and was under the impression that we don't gain anything from installing GHC and Cabal during Stack job.

summoner-cli/src/Summoner/Template/GitHub.hs
```diff
@@ -121,49 +121,121 @@ gitHubFiles Settings{..} = concat
...
         , "    - name: Build"
         , "      run: |"
-        , "        " <> cabalUpdate
+        , "        " <> cabalConfigure
         , "        " <> cabalBuild
         , ""
         , "    - name: Test"
         , "      run: |"
         , "        " <> cabalTest
+        , ""
```
vrom911, line 176: I think this is redundant 🙂

summoner-cli/src/Summoner/Template/GitHub.hs
```diff
@@ -121,49 +121,121 @@ gitHubFiles Settings{..} = concat
...
+        ]
 
     ghActionsVersions :: [Text]
     ghActionsVersions = map
         (\ghc -> "- " <> quote (showGhcVer ghc))
         settingsTestedVersions
 
+    ghActionsExcludes :: [Text]
+    ghActionsExcludes =
+        let latestVersion = viaNonEmpty last $ sort settingsTestedVersions
```
vrom911, line 224: Here you could sort in descending order. You can use `sortWith Down` for that 🙂 
In this case, you would be able to use `head` instead, and it will be faster.

I wonder if we can use `defaultGhc` here, @chshersh ? It is always the `maxBound`, and currently, it is not possible to create the project without using this GHC 🤔

summoner-cli/src/Summoner/Template/GitHub.hs
```diff
@@ -121,49 +121,121 @@ gitHubFiles Settings{..} = concat
...
+            versionsToExclude = filter (\version -> pure version /= latestVersion) settingsTestedVersions
```
vrom911, line 225: If we would pattern just pattern match on the sorted list from the previous line, you could get the head, and the rest at once, without need to filter and walk the list again 🙂

summoner-cli/src/Summoner/Template/GitHub.hs
```diff
@@ -121,49 +121,121 @@ gitHubFiles Settings{..} = concat
...
+            osesToExclude = ["macOS-latest", "windows-latest"]
+            excludes = do
+                os <- osesToExclude
+                version <- versionsToExclude
+                [  "- os: " <> os
+                 , "  ghc: " <> showGhcVer version
+                 ]
```
vrom911, line 232: Very clever implementation! 👍🏼 🧠

summoner-cli/src/Summoner/Template/GitHub.hs
vrom911: I see that it is needed when both `cabal` and `stack` are used. But I think this case could be handled separately, and we may add an empty line only when necessary 🙂 

E.g.
```haskell
    <> memptyIfFalse (settingsCabal && settingsStack) [""]
```

vrom911 (commented):
Great stuff, thanks so much for working on this. And thanks a lot for your patience 🤗

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: I will fix that

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: I personally like the approach of sorting in descending order. It's short, more efficient that sort plus filter and does not depend on the assumption that `defaultGhc` is always used and the latest in case that changes in the future.

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: If we decide to go with sort in descending order I will do that :) If we decide to go with `defaultGhc` I will get rid of sorting and leave just the `filter` call.

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: Done

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: Thanks :) If in this case clever goes with hard to read then I'm willing to replace. Two alternatives that I considered are list comprehension (I see the benefit of being able to merge the filter part as a guard) and dropping `do` notation.

summoner-cli/src/Summoner/Template/GitHub.hs
chshersh: @vrom911 @gdziadkiewicz Let's proceed with sorting in the `descending` order ⏬ Maybe we won't always use `defaultGhc` 🙂  But it's something to keep in mind for future

chshersh:
Btw, @gdziadkiewicz, are you participating in Hacktoberfest this year? The Hacktoberfest rules has changed, and now it's opt-in for all repositories, but we haven't added the `hacktoberfest` topic to Summoner, as we are very busy with Learn4Haskell. But if you are participating, let us know, so your contributions to Summoner would count towards 4 PRs goal 🙂

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: It turned out that without `filter` I don't need the latest version, just the versions to exclude, so I went with `drop 1` after `sortWith Down`.

summoner-cli/src/Summoner/Template/GitHub.hs
gdziadkiewicz: `filter` removed

gdziadkiewicz:
@chshersh I do participate in Hactoberfest, but I already managed to make 4 contributions :) I implemented the fixes as discussed. @vrom911 could you check if after the changes it's good to go from your perspective?

vrom911 (approved):
Looks fantastic! ✨ 
Thanks a lot for working on this one 💪🏼

PR #531, opened 2021-01-28
[#530] Update to GHC-8.10.3, lts-17

### LINKED ISSUE #530: Update to GHC-8.10.3, lts-17

vrom911:
Resolves #530


summoner-cli/src/Summoner/GhcVer.hs
```diff
@@ -89,7 +89,7 @@ baseVerPvp = \case
     Ghc844  -> Pvp 4 11 1 0
     Ghc865  -> Pvp 4 12 0 0
     Ghc884  -> Pvp 4 13 0 0
-    Ghc8102 -> Pvp 4 14 0 0
+    Ghc8103 -> Pvp 4 14 0 0
```
chshersh, line 92: I just learned that since GHC 8.10.2 the `base` version is actually `4.14.1.0`, not `4.14.0.0`. Well, the silent minor version bump was unexpected 😕 
```suggestion
    Ghc8103 -> Pvp 4 14 1 0
```

summoner-tui/src/Summoner/Tui/Field.hs
```diff
@@ -39,6 +39,7 @@ strField t _ = FormFieldState
     , formFields = []
     , formFieldRenderHelper = renderString
     , formFieldConcat = vBox
+    , formFieldUpdate = flip const
```
chshersh, line 42: I'm still puzzled by the meaning of this function, but what can we do...

summoner-cli/examples/cabal-full/.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           - "8.4.4"
           - "8.6.5"
           - "8.8.4"
-          - "8.10.2"
+          - "8.10.3"
```
chshersh, line 21: It is so nice to see all the updates and refreshments to Summoner 😊 
Thanks for this!

chshersh (approved):
This looks great! Amazing job 👏🏻 8️⃣ 1️⃣ 0️⃣ 3️⃣

summoner-tui/src/Summoner/Tui/Field.hs
vrom911: Me too. But we don't use this functionality (I hope). It is not completely clear if it should be `const` or `flip const`. Do you have an idea here?

While in the first case it doesn't matter (as the type there is `() -> () -> ()`, but in the second change we work with `Bool`s 🤔

chshersh: Yeah, I'm very confused. I tried to investigate but failed. Let's keep the current version then 👍🏻

PR #542, opened 2021-03-13
Replace hub with gh

DavSanchez:
Hi guys!

I have modified Summoner so it uses the official `gh` CLI application from GitHub instead of `hub`, which was already having problems with authorization (see https://github.com/github/hub/issues/2655#issue-752529775).

I have tested the behavior and works properly, creating public and private repositories on GitHub, and also run `cabal test all`. Strangely, Travis CI is failing on my repo due to code in other files that I have not modified, though here it seems that it's passing properly. I don't know what could be the cause 😟 

Let me know what you think.

Regards,
David


summoner-cli/src/Summoner/Project.hs
```diff
@@ -311,10 +311,13 @@ doGithubCommands Settings{..} = do
     runGh :: Text -> IO Bool
     runGh repo =
         True <$ "gh" (["repo", "create", "-y", "-d", settingsDescription, repo]
-             ++ ["--private" | settingsPrivate])  -- Create private repository if asked so
```
DavSanchez: I feel bad for removing this and replacing it for an if-then-else. Looked so elegant! Perhaps you know any other way? I'm not that knowledgeable about Haskell (yet!)

summoner-cli/src/Summoner/Project.hs
```diff
@@ -311,10 +311,13 @@ doGithubCommands Settings{..} = do
...
+             ++ [privateOrPublicFlag settingsPrivate])  -- Create private repository if asked so
              $? pure False
 
     ghHelp :: Text -> IO ()
     ghHelp repo = do
         infoMessage "To finish the process manually you can run the following command:"
-        putTextLn $ "    $ gh repo create -d '" <> settingsDescription <> "' " <> repo <> memptyIfFalse settingsPrivate " --private" <> memptyIfTrue settingsPrivate " --public"
```
DavSanchez: Same as above. This looked so cool 😞

vrom911:
Hey @DavSanchez,

Thanks for your ideas. I haven't looked into code yet, as I would like to discuss the idea first.
This sounds like a huge change. I am still using `hub` without any issues, so I do not have plans on moving to `gh` any time soon due to the lack of necessity. Also, all my tools are tight around `hub`, so I do not have an urge to move everything to something new. 

I also want to highlight that we have not received any issue reports due to the `hub` usage. Note that we also suggest using `ssh` when working with GitHub, so this may resolve your issue.

Anyway, I would like to discuss this idea first before rushing into implementing this big change. Could you please elaborate more on the issue that you are encountering?

DavSanchez:
Hi @vrom911,

Thanks for your answer.  Yes I suppose this can be a pretty major change, I propose it because I stumbled upon this issue while trying to create a project with `summoner` and creating the remote repo at GitHub using HTTPS, and found it was not working unless I created a personal access token and used that instead of the password. It's not an issue with `summoner` per se, but with [`hub` using a deprecated API](https://github.com/github/hub/issues/2655#issuecomment-735836048). I suppose using SSH could allow to skip the linked issue, but since apparently `gh` is also being promoted as a more 'official' CLI for interacting with GitHub, I though it would be a good idea to use it with `summoner` too.

Of course this is just a proposal, if you think `hub` is preferable to `gh` please feel free to close the PR or leave it by for a future time (for example, if `hub` stops to be maintained for whatever reason then this could be reconsidered).

Best regards,
David

Issue #519, opened 2020-11-06
Implement a golden-generator executable


chshersh:
Often, when doing major changes to the generated projects, we need to manually update a lot of files. It would be convenient to have a way to automatically update the content of all generated files, so we can inspect them via `git diff` or during PR and see if there's anything suspicious.

I think this should be implemented as an option to our test-suite stanza, because when you have multiple executables, cabal installs all of them, and we don't want our golden-generators to be installed on user machines near the `summoner` executable itself.

PR #549, opened 2021-09-18
[#519] Implement a golden generator executable

chshersh:
This PR implements `gg` — a golden generator executable. It regenerates and updates on disk the content of all projects, scripts and default TOML config. It is useful when doing some massive boring updates (e.g. adding/removing GHC version support). With `gg` you can just change the main `Summoner` code and just run `crun gg` to update all the projects.

The second commit in this PR shows an example of updating years in all projects using `gg`.

It might be counter-intuitive to use `gg` to generate tests and then test against tests themselves. But the main idea is that you usually don't use `gg` after each change. Golden tests ensure that nothing changes unexpectedly. However, when changing e.g. GHC version you expect changes in golden tests. So you can quickly regenerate them, check the diff to verify that the changes look ok and submit them 🙂 If `gg` generated some unexpected results, you can always clean the mess with `git` because all the projects under the version control :octocat:


summoner-cli/summoner.cabal
```diff
@@ -153,17 +159,21 @@ library
...
   build-depends:       summoner
 
-  ghc-options:         -threaded
-                       -rtsopts
-                       -with-rtsopts=-N
+executable gg
+  import:              exe-options
+  hs-source-dirs:      golden
+  main-is:             Main.hs
+  build-depends:       base
```
chshersh, line 171: I wasn't able to use `common-options` here because GHC outputs warning that `base` is not used in this executable (due to `-Wunused-packages`) 😒 So I don't use `relude` and warnings in this simple small file, just `base`, no ketchup, raw `base`

vrom911 (approved):
Wow, this is huge work! Looks awesome, thank you! 👍🏼

PR #578, opened 2025-03-21
Deprecate hub and replace it with gh

skilly-lily:
Since `hub` is no longer active (no releases since 2020), is no longer an official Github tool, and an official Github CLI tool now exists in `gh`, It might be best to move usage of `hub` to `gh`.

I wrote this for myself to use with the `main` branch of `summoner` (so that I could use it for a modern GHC version), and I thought there might be interest in keeping it here, so I decided to clean it up and submit it upstream.

If this change is wanted, but we don't want to deprecate `hub`, let me know: I can build in `gh` support as a fallback and fix the docs.

One additional change: Github no longer requires a paid plan for private repos, so I also updated that doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ project-name/
 To start using Summoner make sure that you have the following tools installed on your machine:
 
 * [`git`](https://git-scm.com) ⩾ 2.28 – to initialize the GitHub repo.
-* [`hub`](https://github.com/github/hub) – to upload the project to GitHub.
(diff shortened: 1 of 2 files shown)


tomjaguarpaw:
Thanks very much for the contribution! However, `summoner` seems to be basically in emergency maintenance only mode. If you're a regular user and want to keep it updated perhaps you could petition @vrom911 to be added as a co-maintainer?

vrom911 (approved):

